PR #19, opened 2021-11-16
Reuse table from napari skimage regionprops instead of copying its code

haesleinhuepf:
Hi Laura @lazigu ,

here comes a PR for the issues with tables we had. E.g. the table column headers are now longer within the table but rather real column headers. With this, also selecting a cell and highlighting it in the image should work again. To achieve this, your plugin will use the table within napari-skimage-regionprops. And if we see issues again, we only need to fix it in one place (there) :-)

Let me know what you think!

Best,
Robert

### DIFF
--- a/napari_clusters_plotter/_utilities.py
+++ b/napari_clusters_plotter/_utilities.py
@@ -13,77 +13,8 @@ def widgets_inactive(*widgets, active):
 
 
 def show_table(viewer, labels_layer):
-    dock_widget = _table_to_widget(labels_layer.properties, labels_layer)
-    viewer.window.add_dock_widget(dock_widget, name='Region properties table', area='right')
-
-
-# adapted from Robert Haase napari-skimage-regionprops
-def _table_to_widget(table: dict, labels_layer: napari.layers.Labels) -> QWidget:
-    """
-    Takes a table given as dictionary with strings as keys and numeric arrays as values and returns a QWidget which
-    contains a QTableWidget with that data.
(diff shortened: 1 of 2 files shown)


haesleinhuepf:
Hi Laura @lazigu ,

could you please have a look at this PR?

Thanks!

Best,
Robert

lazigu:
Hi Robert @haesleinhuepf,

So sorry I've seen it already, and it is a very great solution! But I was hesitant to merge since the plugin is released already and thought maybe I should have more new features implemented before I create a new release. Could you remind me, in this case, if I merge this PR and create a new release, just the last number of version would change, while if there is some new functionality added the middle number changes?

Best,
Laura

haesleinhuepf:
Hi Laura @lazigu ,

I would not worry for making too many small releases. In computer science, there is a way of approaching software projects called: [Release early, release often[(https://en.wikipedia.org/wiki/Release_early,_release_often). It allows developers to collect feedback early and prevents developing things in the wrong direction. ;-)

> just the last number of version would change, while if there is some new functionality added the middle number changes?

Correct. We're just fixing a tiny bug and we expect that the new version is backwards-compatible with the old. Thus, we should change the minor version. If the changes would break backwards-compatibility, we'd change the second. Read more about [semantic versioning](https://semver.org/).

And let me know if you need support with cutting the release. I konw the first times are a bit tricky until one get's used to the procedures... :-)

Best,
Robert

PR #22, opened 2021-12-01
updating the measurement function

Cryaaa:
I edited the measurement function to have the labels start at 1 when loading from file with no column for label number, which should remove any errors when clicking on the table

### DIFF
--- a/napari_clusters_plotter/_measure.py
+++ b/napari_clusters_plotter/_measure.py
@@ -10,6 +10,7 @@ from ._utilities import show_table, widgets_inactive
 from napari_tools_menu import  register_dock_widget
 
 @register_dock_widget(menu="Measurement > Measure intensity, shape and neighbor counts (ncp)")
+
 class MeasureWidget(QWidget):
 
     def __init__(self, napari_viewer):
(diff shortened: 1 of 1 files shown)


napari_clusters_plotter/_measure.py
```diff
@@ -10,6 +10,7 @@
 from napari_tools_menu import  register_dock_widget
 
 @register_dock_widget(menu="Measurement > Measure intensity, shape and neighbor counts (ncp)")
+
```
haesleinhuepf, line 13: Is there a particular reason for adding this empty line here?

haesleinhuepf (approved):
Looks good to me 🌻

napari_clusters_plotter/_measure.py
Cryaaa: not really :D

lazigu:
Thanks, Ryan! This seems not complicated, I could have figured it out when I asked you and complained that it's not working😂

napari_clusters_plotter/_measure.py
lazigu: I will remove the line later anyway because my perfectionist self agrees to let PyCharm reformat files so it removes/adds lines where needed 😂

lazigu (approved):

PR #23, opened 2021-12-01
adding tsne

Cryaaa:
added tsne in a similar way to UMAP, it works for me, maybe somebody else can test it too?


lazigu:
This is great! Thanks so much, Ryan! I will check your PRs in the evening :)

PR #24, opened 2021-12-01
Combining dim red in one widget

Cryaaa:
I was trying to combine both algorithms into one widget so the plugin doesn't get too cluttered with all the different methods. It worked for me, so maybe you can try if it works for you? I think this would be the way to go if it works, but maybe others disagree?

### DIFF
--- a/napari_clusters_plotter/_dimension_reduction.py
+++ b/napari_clusters_plotter/_dimension_reduction.py
@@ -1,0 +1,223 @@
+import pandas as pd
+import warnings
+import napari
+from qtpy.QtWidgets import QWidget, QPushButton, QLabel, QHBoxLayout, QVBoxLayout
+from qtpy.QtWidgets import QListWidget, QListWidgetItem, QAbstractItemView, QComboBox, QSpinBox
+from qtpy.QtCore import QRect
+from ._utilities import widgets_inactive
+
+class dimredWidget(QWidget):
+
+    def __init__(self, napari_viewer):
+        super().__init__()
+
+        self.viewer = napari_viewer
+
+        napari_viewer.layers.selection.events.changed.connect(self._on_selection)
+
+        self.current_annotation = None
+
+        # QVBoxLayout - lines up widgets vertically
+        self.setLayout(QVBoxLayout())
+        label_container = QWidget()
+        label_container.setLayout(QVBoxLayout())
+        label_container.layout().addWidget(QLabel("<b>Dimensionality reduction: UMAP</b>"))
+
+        # selection of labels layer
+        choose_img_container = QWidget()
+        choose_img_container.setLayout(QHBoxLayout())
+        self.label_list = QComboBox()
+        self.update_label_list()
+        self.label_list.currentIndexChanged.connect(self.update_properties_list)
+        choose_img_container.layout().addWidget(QLabel("Labels layer"))
+        choose_img_container.layout().addWidget(self.label_list)
+
+        # selection of dimension reduction algorithm
+        algorithm_container = QWidget()
+        algorithm_container.setLayout(QHBoxLayout())
+        algorithm_container.layout().addWidget(QLabel("Region Properties"))
+        self.algorithm_choice_list = QComboBox()
+        self.algorithm_choice_list.addItems(['   ', 'UMAP', 'TSNE'])
+        algorithm_container.layout().addWidget(self.algorithm_choice_list)
+
+        # selection of n_neighbors - The size of local neighborhood (in terms of number of neighboring sample points)
+        # used for manifold approximation. Larger values result in more global views of the manifold, while smaller
+        # values result in more local data being preserved.
+        self.n_neighbors_container = QWidget()
+        self.n_neighbors_container.setLayout(QHBoxLayout())
+        self.n_neighbors_container.layout().addWidget(QLabel("Number of neighbors"))
+        self.n_neighbors = QSpinBox()
+        self.n_neighbors.setMinimumWidth(40)
+        self.n_neighbors.setMinimum(2)
+        self.n_neighbors.setValue(15)
+        self.n_neighbors_container.layout().addWidget(self.n_neighbors)
+        self.n_neighbors_container.setVisible(False)
+
+        # slection of the level of perplexity. Higher values should be chosen when handling
+        # large datasets
+        self.perplexity_container = QWidget()
+        self.perplexity_container.setLayout(QHBoxLayout())
+        self.perplexity_container.layout().addWidget(QLabel("Perplexity"))
+        self.perplexity = QSpinBox()
+        self.perplexity.setMinimumWidth(40)
+        self.perplexity.setMinimum(1)
+        self.perplexity.setValue(30)
+        self.perplexity_container.layout().addWidget(self.perplexity)
+        self.perplexity_container.setVisible(False)
+
+        # hide widgets unless appropriate options are chosen
+        self.algorithm_choice_list.currentIndexChanged.connect(self.change_nneigbors_list)
+        self.algorithm_choice_list.currentIndexChanged.connect(self.change_perplexity)
+
+        # select properties of which to produce a dimension reduce version
+        choose_properties_container = QWidget()
+        self.properties_list = QListWidget()
+        self.properties_list.setSelectionMode(QAbstractItemView.ExtendedSelection)
+        self.properties_list.setGeometry(QRect(10, 10, 101, 291))
+        self.update_properties_list()
+
+        choose_properties_container.setLayout(QVBoxLayout())
+        choose_properties_container.layout().addWidget(QLabel("Measurements"))
+        choose_properties_container.layout().addWidget(self.properties_list)
+
+        # Run button
+        run_widget = QWidget()
+        run_widget.setLayout(QHBoxLayout())
+        button = QPushButton("Run")
+
+        def run_clicked():
+            if self.get_selected_label() is None:
+                warnings.warn("No labels image was selected!")
+                return
+
+            self.run(
+                self.get_selected_label(),
+                [i.text() for i in self.properties_list.selectedItems()],
+                self.n_neighbors.value(), self.perplexity.value()
+                # todo: enter number of components here as third parameter
+            )
+
+        button.clicked.connect(run_clicked)
+        run_widget.layout().addWidget(button)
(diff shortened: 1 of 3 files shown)


napari_clusters_plotter/_dimension_reduction.py
```diff
@@ -0,0 +1,238 @@
...
+import napari
+from qtpy.QtWidgets import QWidget, QPushButton, QLabel, QHBoxLayout, QVBoxLayout
+from qtpy.QtWidgets import QListWidget, QListWidgetItem, QAbstractItemView, QComboBox, QSpinBox
+from qtpy.QtCore import QRect
+from ._utilities import widgets_inactive
+from napari_tools_menu import  register_dock_widget
+
+@register_dock_widget(menu="Measurement > Dimensionality reduction (ncp)")
+
+class dimredWidget(QWidget):
```
haesleinhuepf, line 12:
```suggestion
class DimensionalityReductionWidget(QWidget):
```
Please use proper classnames. Also Python classes start upper case.

napari_clusters_plotter/_dock_widget.py
```diff
@@ -2,9 +2,11 @@
...
 from ._umap import UMAPWidget
+from ._tsne import TSNEWidget
+from ._dimension_reduction import dimredWidget
 from ._kmeans_clustering import ClusteringWidget
 
 
 @napari_hook_implementation
 def napari_experimental_provide_dock_widget():
-    return [MeasureWidget, PlotterWidget, UMAPWidget, ClusteringWidget]
+    return [MeasureWidget, PlotterWidget, UMAPWidget, ClusteringWidget,TSNEWidget,dimredWidget]
```
haesleinhuepf, line 12:
```suggestion
    return [MeasureWidget, PlotterWidget, UMAPWidget, ClusteringWidget,TSNEWidget,DimensionalityReductionWidget]
```

napari_clusters_plotter/_dock_widget.py
```diff
@@ -2,9 +2,11 @@
 from ._measure import MeasureWidget
 from ._plotter import PlotterWidget
 from ._umap import UMAPWidget
+from ._tsne import TSNEWidget
+from ._dimension_reduction import dimredWidget
```
haesleinhuepf, line 6:
```suggestion
from ._dimension_reduction import DimensionalityReductionWidget
```

napari_clusters_plotter/_dimension_reduction.py
```diff
@@ -0,0 +1,238 @@
...
+        self.label_list = QComboBox()
+        self.update_label_list()
+        self.label_list.currentIndexChanged.connect(self.update_properties_list)
+        choose_img_container.layout().addWidget(QLabel("Labels layer"))
+        choose_img_container.layout().addWidget(self.label_list)
+
+        # selection of dimension reduction algorithm
+        algorithm_container = QWidget()
+        algorithm_container.setLayout(QHBoxLayout())
+        algorithm_container.layout().addWidget(QLabel("Dimension Reduction Algorithm"))
```
haesleinhuepf, line 41:
```suggestion
        algorithm_container.layout().addWidget(QLabel("Dimensionality Reduction Algorithm"))
```

napari_clusters_plotter/_dimension_reduction.py
```diff
@@ -0,0 +1,238 @@
...
+
+            # write result back to properties
+            for i in range(0, n_components):
+                properties["TSNE_" + str(i)] = embedding[:,i]
+
+        else:
+            warnings.warn('No Dimension Reduction Algorithm Chosen!')
+
+
+
```
haesleinhuepf, line 201:
```suggestion
```

napari_clusters_plotter/_tsne.py
```diff
@@ -0,0 +1,170 @@
...
+
+        napari_viewer.layers.selection.events.changed.connect(self._on_selection)
+
+        self.current_annotation = None
+
+        # QVBoxLayout - lines up widgets vertically
+        self.setLayout(QVBoxLayout())
+        label_container = QWidget()
+        label_container.setLayout(QVBoxLayout())
+        label_container.layout().addWidget(QLabel("<b>Dimensionality reduction: TSNE</b>"))
```
haesleinhuepf, line 26:
```suggestion
        label_container.layout().addWidget(QLabel("<b>Dimensionality reduction: t-SNE</b>"))
```
I think the algorithm is spelled like that, right?

napari_clusters_plotter/_tsne.py
```diff
@@ -0,0 +1,170 @@
+import pandas as pd
+import warnings
+import napari
+from qtpy.QtWidgets import QWidget, QPushButton, QLabel, QHBoxLayout, QVBoxLayout
+from qtpy.QtWidgets import QListWidget, QListWidgetItem, QAbstractItemView, QComboBox, QSpinBox
+from qtpy.QtCore import QRect
+from napari_tools_menu import  register_dock_widget
+
+@register_dock_widget(menu="Measurement > TSNE dimensionality reduction (ncp)")
```
haesleinhuepf, line 9:
```suggestion
@register_dock_widget(menu="Measurement > t-SNE dimensionality reduction (ncp)")
```

napari_clusters_plotter/_dimension_reduction.py
```diff
@@ -0,0 +1,238 @@
...
+        self.label_list.currentIndexChanged.connect(self.update_properties_list)
+        choose_img_container.layout().addWidget(QLabel("Labels layer"))
+        choose_img_container.layout().addWidget(self.label_list)
+
+        # selection of dimension reduction algorithm
+        algorithm_container = QWidget()
+        algorithm_container.setLayout(QHBoxLayout())
+        algorithm_container.layout().addWidget(QLabel("Dimension Reduction Algorithm"))
+        self.algorithm_choice_list = QComboBox()
+        self.algorithm_choice_list.addItems(['   ', 'UMAP', 'TSNE'])
```
haesleinhuepf, line 43:
```suggestion
        self.algorithm_choice_list.addItems(['   ', 'UMAP', 't-SNE'])
```

haesleinhuepf (commented):
Hi Ryan @Cryaaa ,

it looks very good to me. I also tried and it works like a charm. I added some suggestions to the code. Please check if it still works after considering my suggestions.

Furthermore, when running the TSNE, I had this warning popping up in napari:
> The PCA initialization in TSNE will change to have the standard deviation of PC1 equal to 1e-4 in 1.2. This will ensure better convergence.

Is there anything we can do about it?

Cryaaa:
@haesleinhuepf Nice one! I'll test if the changes broke anything now and try and get rid of the warning.

haesleinhuepf:
@Cryaaa: In the future, better leave merging to Laura @lazigu . She's the boss here :-)

Cryaaa:
@haesleinhuepf @lazigu Oh, sorry I didn't realise it was going to merge into main. I'll definitely leave it to laura in the future... I thought I branched it off another branch I generated and that it would merge into that branch (If that makes any sense). I also looked at the error and it seems that the code in scikit learn spits out that message automatically if the initialisation is chosen as PCA, so until they remove the message it will always be displayed. This is slightly annoying as PCA is the initialisation method they suggest to use :D

zoccoler:
> @haesleinhuepf @lazigu Oh, sorry I didn't realise it was going to merge into main. I'll definitely leave it to laura in the future... I thought I branched it off another branch I generated and that it would merge into that branch (If that makes any sense).

@Cryaaa It does to me! I had the exact same problem and reasoning you did. I was writing the new entry to the blog. I forked the repository, had to update my "local" main branch to solve differences before a PR and then I applied a merge thinking I was doing it locally, but it actually applied to the main repository...

> I also looked at the error and it seems that the code in scikit learn spits out that message automatically if the initialisation is chosen as PCA, so until they remove the message it will always be displayed. 

Did you try to remove all warnings before that? May be a little too extreme, but I would expect not to see that warning if you put this before that:

```
import warnings
warnings.filterwarnings("ignore")
```
It seems it is a `FutureWarning` [category](https://docs.python.org/3/library/warnings.html#warning-categories) as dislayed [here](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/manifold/_t_sne.py), so you could also potentially filter it by its type, similar to [this](https://docs.python.org/3/library/warnings.html#warning-categories).

zoccoler:
Actually this seems to work:
`warnings.filterwarnings(action='ignore', category=FutureWarning, module='sklearn')`
Adapted from [here](https://stackoverflow.com/a/63980460/11885372)

lazigu:
@Cryaaa no worries, I would have merged it anyway 😀 I really like this solution of having both options in the same widget. I will just change the file name to dimensionality_reduction and remove UMAP from the widget label.

Just for the future, it would super nice if you changed _umap.py directly instead of creating a new file, so not all file would appear green, it would be easier to review the code 😀

@zoccoler Thanks a lot! I will try this out.

Cryaaa:
@lazigu actually the Widget is already called dimensionality reduction widget in the main branch. Additionally there are two widgets: one for umap and one for tsne. I thought we should leave them for now maybe but ultimately we could delete them. Hope that clears things up

lazigu:
Oh sorry, I think I got confused because I just deleted them before reading your message 🙈 Why do you want to leave them if both methods are implemented in dim_red widget? It's a bit confusing if there are too many widgets, no?

Cryaaa:
@lazigu yeah that makes sense, I just didn’t want to delete things without asking. I feel more comfortable just adding new files but I must admit it didn’t make sense 😄

lazigu:
I just realized that I forgot to push my changes, but now I did that so you can see minor changes that I did.
Now I see what you meant, that to change a file name old file is deleted, my mistake, nevermind my comment about green code then 😀
Thanks a lot for this again 🤩
P.s. Marcelo's solution worked!
P.p.s @Cryaaa will you be working now on more contributions? Because I'm working now to simplify a lot of parts of the code using magicgui, so I can already see a lot of conflicts in the future if PRs keep coming 😂

Cryaaa:
@lazigu ahhhhh I see that could indeed be a problem. I was going to work on a few other things but let’s talk about it tomorrow after the group meeting if you have some time?

lazigu:
@Cryaaa yes, sure! Goodnight 😀

Issue #20, opened 2021-11-17
Use things from magicgui for simplify code


Czaki:
using parts of `magicgui` may simplify the codebase:
 * `magicgui.widgets.create_widget` could allow to simplify getting layer selector. `create_widget(annotation=Labels, label="ROI")`  which returns magicgui widget but its `native` property holds Qt Widget and its `value` property holds layer.  This allows not to keep and maintain code for layer selection. Example:  https://github.com/4DNucleome/PartSeg/blob/dedd823837e419839328f5526aec350f25934502/package/PartSeg/plugins/napari_widgets/mask_create_widget.py#L36-L37. To get a refresh of layer list there is a need to define the `reset_choices` function in widget like here: https://github.com/4DNucleome/PartSeg/blob/dedd823837e419839328f5526aec350f25934502/package/PartSeg/plugins/napari_widgets/mask_create_widget.py#L64-L70
 * Using enum for statical combo box like here https://github.com/BiAPoL/napari-clusters-plotter/blob/5e90688096e5dcb82e65c4ebaec40de2ec41857c/napari_clusters_plotter/_measure.py#L48-L51 Enum could be converted to Combobox using `create_widget` from `magicgui` or using QEnumComboBox from `superqt` https://github.com/napari/superqt/blob/main/docs/combobox.md `Enum` allow to overwrite `__str__` and there is IntEnum which allows using and operation to mimic  operations like: https://github.com/BiAPoL/napari-clusters-plotter/blob/5e90688096e5dcb82e65c4ebaec40de2ec41857c/napari_clusters_plotter/_measure.py#L201

Issue #6, opened 2021-11-09
Clustering id selection bug


lazigu:
After closing the plotter widget, and reopening it after kmeans clustering, the plot from the previous session is still plotted and previously selected layers/axes are selected. Therefore, newly measured KMEANS_cluster_id is not visible in the list. The same label image needs to be “reselected” to be able to see KMEANS_cluster_id in the list, which might be confusing.

PR #25, opened 2021-12-02
Simplify the code using magicgui, two new buttons

lazigu:
Hi all!

So this is a new code with some changes. I would love to get some feedback.

**So what's new?**
- Magicgui is used for the selection of labels layer (ComboBox), selection of clustering method (ComboBox), and clustering options (SpinBoxes) (issue #20). I think it could be simplified that instead of creating every time a new Qt widget (container) with a horizontal layout to add a label, a label would somehow be added automatically? This is not an issue of course, because it all works as it is, it would just shorten the code. 
- I added "Restore Defaults" button. For now, it restores kmeans options to default values (nr of clusters and iterations). I'm not sure though if it makes more sense if everything is restored (like clustering method could also be unselected)?
- I added "Update Measurements" button. This could actually be another solution besides the tool menu for #6. The button is only needed when you switch back from the measuring widget, and without having to 'reselect' the same labels layer, measurements list could be updated.
- Some variable names were made prettier, and the file was reformated to make the code nicer.

image

I will be waiting for your thoughts and feedback if it all runs fine and nothing crashes!


napari_clusters_plotter/_kmeans_clustering.py
```diff
@@ -1,158 +1,155 @@
...
-                self.kmeans_nr_iter.value()
+                self.kmeans_nr_clusters.value,
+                self.kmeans_nr_iterations.value
             )
 
-        button.clicked.connect(run_clicked)
-        run_widget.layout().addWidget(button)
+        run_button.clicked.connect(run_clicked)
+        update_button.clicked.connect(self.update_properties_list)
+        defaults_button.clicked.connect(partial(restore_defaults, self, DEFAULTS))
```
haesleinhuepf, line 133: I'm just thinking: If we had defaults, also for other algorithms such as GMM or HDBScan, would "DEFAULTS" still work? Or should we have "KMEANS_DEFAULTS"?

napari_clusters_plotter/_kmeans_clustering.py
```diff
@@ -1,158 +1,155 @@
...
-                    selected_index = i
-                self.label_list.addItem(layer.name)
-                i = i + 1
-        self.label_list.setCurrentIndex(selected_index)
+        widgets_inactive(self.kmeans_settings_container_nr, self.kmeans_settings_container_iter,
+                         active=self.clust_method_choice_list.current_choice == 'KMeans')
 
     def update_properties_list(self):
-        selected_layer = self.get_selected_label()
+        selected_layer = self.labels_select.value
```
haesleinhuepf, line 152: so cool 😎

napari_clusters_plotter/_kmeans_clustering.py
```diff
@@ -161,45 +158,48 @@ def update_properties_list(self):
...
     # this function runs after the run button is clicked
     def run(self, labels_layer, selected_measurements_list, num_clusters, num_iterations):
-        print("Dimensionality reduction running")
-        print(labels_layer)
-        print(selected_measurements_list)
+        print("Selected labels layer: " + str(labels_layer))
+        print("Selected measurements: " + str(selected_measurements_list))
 
-        # turn properties from layer into a dataframe
+        # turn properties from layer into pandas dataframe
```
haesleinhuepf, line 178:
```suggestion
        # turn properties from layer into a pandas dataframe
```

haesleinhuepf (approved):
Hi Laura @lazigu ,

I really like the new layout! It's cleaner and appears more organized. Also the code :-) Great work! A minor request: Can we have the "Restore Defaults" button also for Dimensionality reduction? That would be cool :-)
image

Btw. if you want to make suggestions to others in pull-requests (or to yourself), you can use the suggestion button in the review dialog (I'm just posting this here, because in another channel someone asked about that recently):

image

One more hint; If you send a pull-request as base for discussion and not really for merging (yet), you can mark it as "PR-draft":

image

Have a great weekend!

Best,
Robert

napari_clusters_plotter/_kmeans_clustering.py
Cryaaa: If I read the code correctly the restoring to defaults will go through all widgets in the dictionary and change the values to the default ones. If the HDBSCAN values are also in the dictionary it should restore them simultaneously. Maybe @lazigu can confirm or deny this..

lazigu: It would! I actually tried to implement HDBScan in another branch (it does not yet function properly, but buttons are there already), so you can see that it works to restore defaults :) So my idea is that I will have for different widgets of the plugin different Defaults dictionaries, and then I am passing it to restore_defaults function in the utilities.

napari_clusters_plotter/_kmeans_clustering.py
lazigu: all thanks to the smart person who gave me tips in issues section :)

lazigu:
Hi Robert @haesleinhuepf ,
So cool, I didn't know about PR as a draft option, thanks!! And thanks for showing the suggestions button again, I forgot about it 😉

And yes I am planning to make the defaults button for the other widget also. 
Thanks for the feedback!!

Cryaaa (commented):
I just wanted to say great job, it looks like implementing magicgui really makes some things much easier, but to be honest I have not fully grasped  how everything works. I also wanted to say that the added functionality works for me and I think the update measurements button makes a lot of sense!

napari_clusters_plotter/_kmeans_clustering.py
Cryaaa: That sounds great! if you need any help with HDBSCAN just let me know, maybe I have already stumbled into the same problem :)

lazigu: Ah I didn't see this message from you but I already took up on your offer :D

PR #26, opened 2021-12-03
Add hdbscan and standard scaler option

lazigu:
Hi Ryan @Cryaaa,

I tried to implement **HDBSCAN** but it is not yet working, I got lost in the documentation, maybe for you it will be way faster to implement it. Could you take a look, please? It is almost done I think, but I got confused at which point results that we could put in the regprops table are actually returned.

I also tried to enable the **standard scaler option** before both clustering methods. It would be nice if you checked if it works as it should.

**Couple more changes in this branch:**

- CLUSTER_ID is saved as CLUSTER_ID_SCALER_BOOL. I thought it is nice in some cases to perform clustering with standard scaling and without and then maybe compare plots since both results would be saved in properties of the labels layer.
- Measurements box now does not include "labels" and "CLUSTER_ID" options.
- Added a couple of more error checkings and warnings.

I would love somehow to implement that when you hover with a mouse over some parameters (like min number of samples), an explanation box would appear, so not experienced in clustering users could find out more what each parameter means.

And, Ryan, please feel free to directly commit to this branch if you work on clustering.py, and my next steps will be to tidy up (and use magicgui where appropriate) for the other widgets :)


requirements.txt
```diff
@@ -19,4 +19,5 @@ pyqt5~=5.15.4
 qtpy~=1.11.2
 scikit-image~=0.18.3
 umap-learn~=0.5.1
-scikit-learn~=0.24.2
\ No newline at end of file
+scikit-learn~=0.24.2
+hdbscan~=0.8.27
```
Cryaaa, line 23: @lazigu I had to add hdbscan to the requirements because it didn't install automatically (I got an error when trying to run the hdscan code for the first time). I'm not sure if this is enough though?

lazigu: Hmm I thought that I messed up hdbscan clustering function and it was not returning what I expected as in kmeans function for example. But mayyybe it was related to installation. Strange that it was not working automatically.

Cryaaa: Oh so the function was also not working because you used the wrong function. Instead of generate_prediction_data() you should have just used fit_predict(measurement) just like in kmeans

lazigu: Ahhhh many thanks!! P.s. requirements.txt is a list for developers, while setup.cfg has a list what will be installed when using pip . So if hbdscan is there, strange that it’s not installed automatically (I can’t check now, I’m on a phone 🙈)

Cryaaa: It wasn't there before so thats probably why it didn't work but I haven't checked yet either and I'm trying to fix something else at the moment :D

lazigu: Ok no hurrying, I will look into it later!

PR #32, opened 2021-12-09
Fixing loading csv bug

Cryaaa:
Fixed bug where loading regionproperties with a label column broke clustering. Only a small change so I think it should be fine to merge


lazigu (approved):
Thanks, Ryan! Yes, indeed this seems like a small bug fix so I will merge right away.

PR #33, opened 2021-12-09
Add questions marks and tooltips

lazigu:
Question marks added near min_clusters_size and min_nr_samples, which display short descriptions when hovered with a mouse and also redirect to a page with more info when clicked.
image

@Cryaaa, any ideas near which variables such feature would also be useful? I thought for kmeans parameters it's pretty self-explanatory so I didn't put them there.


Cryaaa:
@lazigu 
That is so cool! I really like that addition! So definitely there should be some in the dimension reduction widget for the number of neighbors parameter and perplexity parameter and possibly also one where you choose the dimension reduction algorithm with some hints of how the algorithms are different from each other. I could try and write some texts for those question marks. 
I definitely think that more will be necessary when further algorithms are implemented

lazigu:
Yes I also think for those parameters it would make sense. So if you want/have time to do this, you can also add some, otherwise, I will do it a bit later :)

PR #34, opened 2021-12-10
Update toolbar

### LINKED ISSUE #28: Manual cluster selection after zooming in plot
When I zoom in on the plot I cannot get back to the lasso selector for selecting manual clusters. I guess a button for that would be useful or that it reverts back to it automatically after zooming in.

zoccoler:
This should fix #28 , at least give the user visual feedback.

- updated Zoom and Pan buttons' tooltip (to explain these are toggle buttons, you activate and deactivate them);
- changed toolbar icons to white (had to add icon images in folder, they were edited (color inverted) from matplotlib original icon images);
- updates icons when highlighted to give user visual feedback that they are selected.

P.S. matplotlib generates these icons from a pixelmap and a mask by means of an internal function. I could not modify an internal parameter of that function, that's why I am uploading white icons in large format. Result should be the same.


lazigu:
Hi Marcelo @zoccoler,

I somehow do not see any icons when I try to run napari. I tried to figure out maybe something is wrong with paths of images files, but I couldn't make it work. Any idea what could be the problem?

image

napari_clusters_plotter/_plotter.py
```diff
@@ -146,7 +147,21 @@ class MyNavigationToolbar(NavigationToolbar):
...
         super().__init__(canvas, parent)
         self.canvas = canvas
-
+    
+    def _update_buttons_checked(self):
+        super()._update_buttons_checked()
+        # changes pan/zoom icons depending on state (checked or not)
+        if 'pan' in self._actions:
+            if self._actions['pan'].isChecked():
+                self._actions['pan'].setIcon(QIcon("images//my_toolbar_icons//pan_checked.png"))
```
lazigu, line 156:
```suggestion
                self._actions['pan'].setIcon(QIcon("images//my_toolbar_icons//Pan_checked.png"))
```
Should this maybe be a capital letter since file names are starting with capital letters, or it doesn't matter? This didn't fix the problem that I still don't see the icons though

zoccoler:
> Hi Marcelo @zoccoler,
> 
> I somehow do not see any icons when I try to run napari. I tried to figure out maybe something is wrong with paths of images files, but I couldn't make it work. Any idea what could be the problem?
> 
> image

Hmm it seems a path issue. Do you see the new icons in the new folder I created (images/my_toolbar_icons) ?

Tha capital letters could be part of the problem, but that's not the part where the items are loaded (I fixed it anyway, thanks!)

I am going to be more strict about the path in this next commit (I let some 'prints' for now just so we can double check the path of the icons; we'll delete them after if it works). Could you try it again? Also please let me know if you see the new folder and the printed paths. Thx 😄

napari_clusters_plotter/_plotter.py
zoccoler: It could contribute to the problem, but it's not its main cause. Thanks for noticing that, I updated it in the next commit 😃

napari_clusters_plotter/_plotter.py
```diff
@@ -146,7 +148,21 @@ class MyNavigationToolbar(NavigationToolbar):
...
         super().__init__(canvas, parent)
         self.canvas = canvas
-
+    
+    def _update_buttons_checked(self):
+        super()._update_buttons_checked()
+        # changes pan/zoom icons depending on state (checked or not)
+        if 'pan' in self._actions:
+            if self._actions['pan'].isChecked():
+                self._actions['pan'].setIcon(QIcon(os.path.join(os.getcwd(), "images", "my_toolbar_icons", "Pan_checked.png")))
```
haesleinhuepf, line 157: Hi Marcelo @zoccoler ,

the icons should live within the plugin directory (e.g. "napari_clusters_plotter\icons") so that they are deployed together with the code to the end-user. Btw. `os.getcwd()` may point to the folder from which napari was started. Consider using `__file__` like [here](https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/master/napari_pyclesperanto_assistant/_gui/_button_grid.py#L6) which points to the folder when the current python file lives. 🌻

Best,
Robert

zoccoler: Hi Robert @haesleinhuepf ,

Thanks, I had used relative paths previously, but your suggestion of taking the path from `__file__` is indeed a more accurate solution. I have implemented that now.
About the icons, I had put them inside the plugin at 'images\my_toolbar_icons' before. Nevertheless, I have moved them to this new directory 'napari_clusters_plotter\icons' for simplicity and perhaps as being a more appropriate location.

Obs: left some prints, we should remove them before merging if this works.

lazigu:
Sorry for the delayed reply. I can see icons now, but something is wrong because manual clustering does not work (nothing happens after lasso selection)🤔 Does it work for you?
image

P.s. I also realized that we should have a way to deal if napari is used in the light version, but I guess that's the task for the future.

zoccoler:
@lazigu it doesn't work here either, but I didn't touch that part of the code.

I noticed the problem might be my Github desktop is not updating properly, this is still an [open issue there](https://github.com/desktop/desktop/issues/11334#issuecomment-756627312), unless I get beta version.

I will try to fix this, the error is probably because my PR is over an outdated version of the code.

zoccoler:
I think I found the problem. Two packages had the same name...
I added `Path` from `pathlib`, but there was already a `Path` from `matplotlib.path`.

Now I changed the former to `PathL`. @lazigu could you check again now? 😄

lazigu:
Works like a charm now, thanks! 😄

lazigu:
Works like a charm now, thanks! 😄

lazigu:
Works like a charm now, thanks! 😄

PR #38, opened 2022-01-18
Use layer features instead of layer properties

### LINKED ISSUE #35: Selecting data points doesn't work anymore in ncp 0.2.1
Hi Laura @lazigu, hi all,

I was just excited about the new version and tested it. Unfortunately, selecting data points in the plot doesn't work anymore. :-(

Here is what I did:
* Install napari, napari-clusters-plotter, napari-segment-blobs-and-things-with-membranes, napari-pyclesperanto-assistant
* I opened `File > open examles > clesperanto > blobs (from ImageJ)`
* I segmented it using `Tools > Segmentation / labeling > Voronoi Otsu Labeling (nsbatwm)` using paramerters spot_sigma=4 and outline_sigma=2.
* I measured the properties of these blobs using napari-clusters-plotter`Measure now (with neighborhood data)`
* I started the plotter from the menu `Tools > Measurement > Plot Measurements (ncp)` and selected `area` versus `mean intensity` as shown in this video:

https://user-images.githubusercontent.com/12660498/147076112-9108ec36-8de1-4265-b76b-5d8830cc0bd1.mp4

As you can see, the selection of data points doesn't work anymore. Let me know if there is anything I can do to help fixing it. :-)

Best,
Robert

lazigu:
Hi all,

This PR fixes #35.
It also makes the plugin incompatible with napari < 0.4.13, so I added the comment in the documentation.
I will be waiting for your feedback.

Best wishes,
Laura

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,12 +97,11 @@ Afterwards, you can again save and/or close the table. Also, close the clusterin
 
 ### Plotting clustering results
 Return to the Plotter widget using the menu `Tools > Measurement > Plot measurement (ncp)`.
-Select `UMAP_0` and `UMAP_1` as X- and Y-axis and the `ALGORITHM_NAME_CLUSTERING_ID` as `Clustering`, and click on `Run`.
+Select `UMAP_0` and `UMAP_1` as X- and Y-axis and the `ALGORITHM_NAME_CLUSTERING_ID` as `Cluster`, and click on `Run`.
 
 ![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/hdbscan_clusters_plot.png)
 
 Example of k-means clustering results:
-
 ![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/kmeans_clusters_plot.png)
 
 ## Installation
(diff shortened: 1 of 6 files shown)


haesleinhuepf (commented):
Hey Laura @lazigu ,

let's wait a second before we tear down the bridges behind us. ;-) Would it be possible to be compatible to both `properties` and `features` for some transition time? I'm not sure if all users will immediately jump on napari==0.4.13... maybe we can program little functions (pseudocode) like this:

```
def set_features(layer, tabular_data):
    if hasattr(layer, "properties"):
        layer.properties = tabular_data
    if hasattr(layer, "features"):
        layer.features = tabular_data

def get_tabular_data(layer):
    if hasattr(layer, "features") and layer.features is not None:
        return layer.features
    if hasattr(layer, "properties") and layer.properties is not None:
        return layer.properties
    return None
```

What do you think?

lazigu:
Good idea! I will do it tomorrow :)

lazigu:
Hi Robert @haesleinhuepf, this is ready for review again, I tested and it worked for me with both napari versions.
Thanks a lot for the functions above, made the job very easy :)

haesleinhuepf (approved):
Hi Laura @lazigu ,

that looks great! I also like `add_column_to_layer_tabular_data`. This trick of "making a function out of something we used in multiple places" is very good, especially when dependent libraries (napari in our case) change something. Assume, they rename "features" to "measurements" with the next release. We would then have a super-easy fix to implement, ;-)

All the changes look very good to me. Well done!

PR #42, opened 2022-02-01
add DOI badge

haesleinhuepf:
Hi Laura @lazigu ,

this PR just adds a DOI badge to the readme. It links to this page:

https://doi.org/10.5281/zenodo.5884658

Which can be used to scientifically cite the software. I configured Zenodo so that it automatically updates that page everytime we create a new release. I'm happy to show you (or anybody else) how this works. It is surprisingly easy:

https://genr.eu/wp/cite/

Best,
Robert


lazigu:
Hi Robert, this looks great, thanks! And I would definitely like to see how this works.

Issue #36, opened 2022-01-12
Progress bar for k-Means clustering and friends


haesleinhuepf:
Hi all,

Fabrice Senger had a good idea for improving ncp: A progress bar for the operations that take a long time:
https://forum.image.sc/t/dont-yet-support-nd-fancy-indexing/61878/2

There's something about this in the napari documentation:
https://napari.org/docs/dev/api/napari.utils.progress.html

Happy to have a chat about that request. It's not urgent btw. but I think nice-to-have. :-)

Best,
Robert

lazigu:
Hi @haesleinhuepf,

I actually also had this idea, but because for me napari usually stops responding even during not demanding tasks, and 'comes back to life' soon, I gave up on the idea :D
But I understand that for other users it would be great to have this implemented so I will try!

Best wishes,
Laura

PR #44, opened 2022-02-02
Implement progress bar, multithreading

lazigu:
Hi all,

I implemented multithreading in the dimensionality reduction widget, so now napari does not crash during the computation. I've tried also to implement a progress bar (#36). However, my long-running function (umap or tsne) are not generators, therefore, the progress bar that I get is only an indeterminate progress bar (video below).

https://user-images.githubusercontent.com/52177660/152180924-b0d4281a-362a-4b6e-986f-05023f1cdf9c.mp4

Umap itself is also able to provide a tqdm progress bar in the console:

image

However, since umap is running in a secondary thread I was googling is it possible to somehow make it that the main thread receives this umap tqdm and it is displayed in napari itself not the console. I found that "PyQt widgets cannot be modified from any thread other than the application thread, and since the updates to stdout and stderr are coming from the secondary worker thread it would violate that". This is from this stackoverflow thread, which I thought could help me, but I cannot grasp the code there:
https://stackoverflow.com/questions/21071448/redirecting-stdout-and-stderr-to-a-pyqt4-qtextedit-from-a-secondary-thread

I thought I could somehow get the umap tqdm to output to another widget I create instead of relying on napari's progress bar, which is not able to display actual progress because the worker is a function, not a generator, so nothing is yielded. Does it seem doable to you? Or do you think it's too much struggle and it's enough to have this progress indication as it is now?

I know it is possible to have two-way multithreading, where it is possible to send and receive values to and from the main thread, but then again a generator is needed:
https://github.com/napari/napari/blob/6a3e11aa717e7928a0a5a3c7693577729a466ef1/examples/multithreading_two_way.py 

It seems I cannot figure this out on my own. @jo-mueller, I think you are quite an expert with tqdm, maybe you will be able to help me out? @haesleinhuepf, thoughts?

Thanks a lot for any feedback!

Best,
Laura


jo-mueller:
Hey Laura @lazigu ,

> I found that "PyQt widgets cannot be modified from any thread other than the application thread

I think this may be only partially true. You can install signals to your threads which can be "emitted" by the thread and conected to from the outside of the thread. `button.clicked` is an example of this system. There is a [tutorial ](https://napari.org/guides/stable/threading.html)on the napari page that explains how to do this (see at section: "Adding custom signals".)

> and since the updates to stdout and stderr are coming from the secondary worker thread

I played around with that feature in a different context, but, retrospectively, it's more trouble that it's worth because while rerouting stdout (all print statements, tqdms, etc) to a custom output (like a custom-made progressbar) can be nice for visualization, it also mutes the IPython console which can make debugging incredibly difficult :/

>  the worker is a function, not a generator, so nothing is yielded
This may be the key problem here...do I understand it correctly that the umap function iterates over something (internally), which you would like to display somehow? I fear that it may be super difficult to actually catch and display the progress happening within umap. I think it may be wise to leave it with a sign that shows that the plugin is working and not frozen.

One thing I could think of: If the processing time of umap/other alogorithms scales in a predictable way with time to simply print out an estimate on how long it could take. And maybe a stop button if the result ends up being multiple hours or so ^^

lazigu:
> I played around with that feature in a different context, but, retrospectively, it's more trouble that it's worth because while rerouting stdout (all print statements, tqdms, etc) to a custom output (like a custom-made progressbar) can be nice for visualization, it also mutes the IPython console which can make debugging incredibly difficult :/

Ah ok, so it really seems not worth it then.

> One thing I could think of: If the processing time of umap/other alogorithms scales in a predictable way with time to simply print out an estimate on how long it could take. And maybe a stop button if the result ends up being multiple hours or so ^^

This is a good idea, I will look into it. Ryan @Cryaaa, by any chance maybe you have encountered estimating how long would umap take during your master's work?

Thanks for the help @jo-mueller!

lazigu:
This is finished and ready for the review now. I left only napari progress bar, and multithreading was implemented in all widgets except plotting one (possible issue for the future).
This branch contains merged add_clustering_methods branch already, so it should be merged after #65, #66 and #67 are already merged, and released in the separate release in case of possible bugs. I will request review after other PRs are tackled.
It would also close #68.

P.s. Tests on [macos-latest py3.9](https://github.com/BiAPoL/napari-clusters-plotter/runs/5759565574?check_suite_focus=true#logs) run over 20 min now, it should be taken a look if it does not finish successfully.

lazigu:
Hey @jo-mueller,
I am getting the same error, but only for the call_to_function test, not test_umap. I was running tests just by typing in pytest:

image

Or also with coverage: 
`coverage run --source=napari-clusters-plotter -m pytest -v napari_clusters_plotter && coverage report -m`

That´s how it looks like:
image

test_umap also passes if I run it directly in pycharm.

PR #45, opened 2022-02-17
Moving measurement functions outside of the class

Cryaaa:
Hey @lazigu! 
This is just a small suggested change to be able to use the measurement functions for timelapse processing. Nothing is really changed and it is based completely on what you had written in the run function!


lazigu:
Thanks @Cryaaa!

PR #46, opened 2022-02-17
Implementing PCA

Cryaaa:
Hey guys,
so I made this branch a while ago, which is most likely why automatic merging is not possible. Maybe you can take a quick look @lazigu if this is a big problem? If it is too complicated I can reimplement it into the current main branch..

### DIFF
--- a/napari_clusters_plotter/_dimensionality_reduction.py
+++ b/napari_clusters_plotter/_dimensionality_reduction.py
@@ -20,8 +20,8 @@ from qtpy.QtWidgets import (
 from ._utilities import (
     add_column_to_layer_tabular_data,
     get_layer_tabular_data,
-    set_features,
     restore_defaults,
+    set_features,
     widgets_inactive,
 )
 
@@ -398,13 +398,14 @@ class DimensionalityReductionWidget(QWidget):
             # reduce dimensionality
             embedding = pca(properties_to_reduce, explained_variance, pca_components)
 
-            # check if principle components are already present 
+            # check if principle components are already present
             # and remove them by overwriting the features
             tabular_data = get_layer_tabular_data(labels_layer)
-            dropkeys = [column for column in tabular_data.keys() if column.startswith("PC_")]
-            df_principal_components_removed = tabular_data.drop(dropkeys, axis = 1)
-            set_features(labels_layer,df_principal_components_removed)
-
+            dropkeys = [
+                column for column in tabular_data.keys() if column.startswith("PC_")
+            ]
+            df_principal_components_removed = tabular_data.drop(dropkeys, axis=1)
+            set_features(labels_layer, df_principal_components_removed)
 
             # write result back to properties
             for i in range(0, len(embedding.T)):
--- a/napari_clusters_plotter/_utilities.py
+++ b/napari_clusters_plotter/_utilities.py
@@ -31,11 +31,13 @@ def get_layer_tabular_data(layer):
         return layer.features
     return None
 
+
 def add_column_to_layer_tabular_data(layer, column_name, data):
     if hasattr(layer, "properties"):
         layer.properties[column_name] = data
     if hasattr(layer, "features"):
         layer.features[column_name] = data
+
 
 def get_nice_colormap():
     colours_w_old_colors = [


Cryaaa:
@lazigu, yeah it was working before... I'll go through the code and see what could be causing the problem :)

lazigu:
@Cryaaa then it will really be a silly mistake I introduced when solving conflicts. I can also look at it tomorrow :)

Cryaaa:
@lazigu unfortunately I cannot reproduce the error right now. I have a relatively fresh environment installed which makes me believe it might be a bug on your side? Could you maybe explain what you did exactly so I can try to reproduce it?
Anyway while trying out different values I realised one more thing we need to add in the code to improve the functionality because as soon as the principal components are added to the table they cannot be removed even when you run it again with different parameters..

lazigu:
@Cryaaa weird but I cannot reproduce it also anymore, it works fine! So before when I encountered the error, I was testing with a drawn a couple of labels by hand because I didn't have saved labels image and some kind of segmentation plugin on hand. But now I tried testing it the same way, and it worked perfectly. I wish I could reproduce because it was really weird

> Anyway while trying out different values I realised one more thing we need to add in the code to improve the functionality because as soon as the principal components are added to the table they cannot be removed even when you run it again with different parameters..

Do you think this is needed? Should the column just be overwritten then like in the case of umap?

In addition, while playing around now I noticed that we should add more error catching that would produce warnings, I noticed such errors (this is not related just to this PR but also in the main branch). Basically, they appear if you play with plugin randomly without paying much attention to your data and parameters selected :D
So a little #todo list:

`ValueError: n_components=4 must be between 0 and min(n_samples, n_features)=2 with svd_solver='full' (related to this PR)`

`ValueError: zero-size array to reduction operation maximum which has no identity
Selected labels layer: Labels
Selected measurements: ['min_intensity', 'max_intensity', 'sum_intensity', 'area', 'mean_intensity', 'centroid_x', 'centroid_y', 'centroid_z', 'mean_distance_to_centroid', 'standard_deviation_intensity', 'max_distance_to_centroid', 'mean_max_distance_to_centroid_ratio']
Dimensionality reduction started (UMAP, standardize: True)...
C:\Users\zigut\Miniconda3\envs\ncp-dev\lib\site-packages\umap\umap_.py:2344: UserWarning: n_neighbors is larger than the dataset size; truncating to X.shape[0] - 1!
C:\Users\zigut\Miniconda3\envs\ncp-dev\lib\site-packages\umap\umap_.py:132: UserWarning: A large number of your vertices were disconnected from the manifold.
Disconnection_distance = inf has removed 0 edges.
It has fully disconnected 2 vertices.
You might consider using find_disconnected_points() to find and remove these points from your data.
Use umap.utils.disconnected_vertices() to identify them.!`

`ValueError: Input contains NaN, infinity or a value too large for dtype('float32').`

Cryaaa:
@lazigu 
So regarding the errors you could produce it seems that most of them are because you have a very small number of cells (probably because you just hand annotated them). To be honest I don't think dimensionality reduction makes sense when you have less than 10 samples to look at so I'm not sure how relevant they are to most use cases but it might make sense to write a custom warning in those cases. 

> Do you think this is needed? Should the column just be overwritten then like in the case of umap?
I do because for some settings you will have 16 principal components and for others 7. When you have the 16 first the first 7 will be overwritten and the others just stay there and I think that that could be misleading.. Removing the principal components from the table before adding the newly calculated ones shouldn't cause any problems though :)

lazigu:
> So regarding the errors you could produce it seems that most of them are because you have a very small number of cells (probably because you just hand annotated them). To be honest I don't think dimensionality reduction makes sense when you have less than 10 samples to look at so I'm not sure how relevant they are to most use cases but it might make sense to write a custom warning in those cases.

Yes yes, that is exactly why these errors appeared, I had just a few blobs labeled. But I am thinking maybe it would be useful to have this in case some users that are playing around with a plugin do not have much knowledge about dimensionality reduction🤔 I will add it, it's just a couple of lines of code.


> I do because for some settings you will have 16 principal components and for others 7. When you have the 16 first the first 7 will be overwritten and the others just stay there and I think that that could be misleading.. Removing the principal components from the table before adding the newly calculated ones shouldn't cause any problems though :)

Ok, this totally makes sense.

Also would be cool to have tooltips explaining parameters that can be selected for PCA.

PR #47, opened 2022-02-17
Feature selection implementation

Cryaaa:
This is the implementation of a feature selection widget which for now only includes the correlation filter from my master thesis. THere seems to be a small issue with the importing of the widget but I don't think that it will cause any greater problems.


lazigu:
pre-commit.ci autofix

lazigu:
@Cryaaa great job with this!!
Just a small suggestion, could you maybe add to the documentation a typical workflow with a screenshot like it is for other widgets? It's a bit confusing for someone who is not familiar with this method e.g. me :D Should Run button maybe be the last one? Also, some tooltips, hovering on the "?" to get more info like in other widgets would be great

Cryaaa:
@lazigu 
That makes sense, I think it could be made simpler possibly but the way I implemented it was a workaround for a problem I found. I really need to clean up the code anyways, so I will add the tooltips at the same time. Maybe we can mark it as a draft PR again until I have finished that.

Cryaaa:
@lazigu,
This PR is also nearing completion now but I think we need to talk about some GUI stuff maybe in a meeting. No rush at all though, since we have other PRs to merge before :). Just wanted to keep this in mind for later!

lazigu:
Closing this due to inactivity. @Cryaaa, feel free to reopen this if you ever have time/want to work on this again, but for now, a bit too many conflicts between the main branch are present.

PR #52, opened 2022-03-01
Napari hub categories

lazigu:
Hi all,

I have added categories that I think are suitable for the clusters-plotter. You can find the choices list [here](https://github.com/edamontology/edam-bioimaging/blob/main/releases/EDAM-bioimaging_alpha06.csv). If anybody wants to review this, and can find some more categories or think some are not suitable, you are more than welcome, otherwise, I would just merge this soon :)


jo-mueller (approved):

Cryaaa (approved):
I don't have any further tags I could find so all clear from my side

PR #53, opened 2022-03-02
Fix tests add codecov

lazigu:
Hi all,

So I finally managed to make that a workflow that runs tests and produces test reports is now successful using GitHub Actions. make_napari_viewer indeed worked. However, I had to comment out all tests that include functions that use cle functions (so Marcelo's @zoccoler  test_processing_dask_array function and Johannes' @jo-mueller test_measurements function), otherwise, it crashes. The error looked like in the picture below. @haesleinhuepf, any idea what am I missing? Or is not possible to use this way GPU accelerated functions, and the only solution is to run those tests just locally?

image


jo-mueller (approved):

zoccoler (commented):
Hi @lazigu,
I am also not sure what causes these errors on Github side.
I agree to let these tests commented for now so that codecov works. But I think we should definitely come back to this later to understand the reason.

If you merge this, could you please create a new issue to remind us about it? :)

lazigu:
> If you merge this, could you please create a new issue to remind us about it? :)

Yes, definitely, I will wait till Monday because Robert said he will take a look then. If there is nothing we can do, I will merge it like it is and create an issue.

haesleinhuepf:
Hi Laura @lazigu , 

as mentioned earlier, also my knowledge about github-CI is limited. I tried to activate tests in #54 , but github doesn't run those tests at all. It might be necessary to merge this PR first and then run the tests on the other PR again. Anyway, I also played with [oclgrind](https://github.com/jrprice/Oclgrind) to replace POCL in [another napari plugin that uses pyopencl](https://github.com/clEsperanto/napari_pyclesperanto_assistant/pull/54) and was able to run tests on MacOS. Windows is still failing. I would not take github CI too serious in this context. One can invest time in fixing GPU-related tests on a system which doesn't have GPUs (i.e.g github-CI), or one invests that time in other work ;-)

lazigu:
@haesleinhuepf thanks for looking at this! :) Yes, the tests on the main branch are not yet running, I fixed that in this PR. 
In general, for me, it is also ok for me to leave these GPU-related tests commented out because after your comment I understand the issue more. I fully agree that one should invest time elsewhere then, I was just thinking that maybe I was missing something simple that could fix the problem.
I will merge this PR then, and later look at PR 54 and the other link you attached :)

PR #57, opened 2022-03-10
Update _dimensionality_reduction.py

Cryaaa:
adding the tooltips for PCA. Maybe you can confirm or deny if they are useful? :D @lazigu


lazigu:
@Cryaaa Very useful!! Thanks :)
I will just fix line-length so pre-commit check stops failing.

PR #65, opened 2022-03-25
Fix nans in measurements list

### LINKED ISSUE #56: Errors due to NaN values if the first label is background
In case segmentation is done other ways than suggested in the documentation, the first label might be a background label. This would cause errors due to the introduced NaN values in the region props table for the background label. For instance, error when trying to do dimensionality reduction.

Therefore, in the measurements widget, a checkbox "the first label is background" could be added, and if the checkbox was checked, the first row could be removed from the region props table.

jo-mueller:
Closes #56 

## Major changes:
- The dimensionality reduction plugin now only takes rows from the measurements values that have no NaNs. This is done with the pandas `DataFrame.dropna()` command. At the same time, the information about which labels are included for this is stored. 
- The `add_column_to_layer_tabular_data()` is changed to handle this. It now receives a target layer into which the new data should be appended and a new dataframe that should be added. The new dataframe must have one column in common with the already available tabular data. In this case, this is the `labels` column. The new values are then sorted into `layer.properties` into the correct rows; the rows with no data from the dimensionality reduction will have a `NaN` entry.
- The clustering procedure has been changed to support appending the cluster index to the data. However, replacing the `NaN` values is not necessary here as the implemented algorithms (kmeans and hdbscan) can apparently handle `NaN` and will give these data points the cluster index `-1`.

## Code style changes
- I moved the dimensionality reduction and the clustering functions out of the widget. Thus, they can be tested easier.
- I added type annotations and default values to the clustering/dimensionality reduction functions as well as default arguments.
- I added docstrings for the clustering functions which could make it easier in the future to also use these functions from code.

### DIFF
--- a/napari_clusters_plotter/_clustering.py
+++ b/napari_clusters_plotter/_clustering.py
@@ -322,58 +322,69 @@ class ClusteringWidget(QWidget):
         self.labels_select.reset_choices(event)
 
     # this function runs after the run button is clicked
-    def run(
-        self,
-        labels_layer: Labels,
-        selected_measurements_list: list,
-        selected_method: str,
-        num_clusters: int,
-        num_iterations: int,
-        standardize: bool,
-        min_cluster_size: int,
-        min_nr_samples: int,
-    ):
-        print("Selected labels layer: " + str(labels_layer))
-        print("Selected measurements: " + str(selected_measurements_list))
-        print("Selected clustering method: " + str(selected_method))
+    def run(self, labels_layer: Labels,
+            selected_measurements_list: list,
+            selected_method: str,
+            **kwargs):
 
-        features = get_layer_tabular_data(labels_layer)
-
-        # only select the columns the user requested and remove NaNs
-        selected_properties = features[selected_measurements_list]
-        non_nan_entries = features.dropna().index
-        non_nan_labels = features['label'].iloc[non_nan_entries]
-
-        # perform clustering
-        if selected_method == "KMeans":
-            y_pred = kmeans_clustering(
-                standardize, selected_properties, num_clusters, num_iterations
-            )
-            print("KMeans predictions finished.")
-
-        elif selected_method == "HDBSCAN":
-            y_pred = hdbscan_clustering(
-                standardize, selected_properties, min_cluster_size, min_nr_samples
-            )
-            print("HDBSCAN predictions finished.")
-        else:
-            warnings.warn(
-                "Clustering unsuccessful. Please re-check selected options."
-            )
-            return
-
-        # write result back to features/properties of the labels layer
-        df_clusters = pd.DataFrame(non_nan_labels, columns = ['label'])
-        df_clusters[
-            f"{selected_method}_CLUSTER_ID_SCALER_{str(standardize)}"
-            ] = y_pred
-
-        add_column_to_layer_tabular_data(labels_layer, y_pred)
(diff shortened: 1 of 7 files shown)


jo-mueller:
@Cryaaa : Would you mind giving this a brief look before I change it to ready for review? Tests are passing but...just to be sure.

napari_clusters_plotter/_clustering.py
```diff
@@ -323,56 +326,112 @@ def reset_choices(self, event=None):
...
+    """
+    print("Selected labels layer: " + str(labels_layer))
+    print("Selected measurements: " + str(selected_measurements_list))
+    print("Selected clustering method: " + str(selected_method))
+
+    features = get_layer_tabular_data(labels_layer)
+
+    # only select the columns the user requested and remove NaNs
+    selected_properties = features[selected_measurements_list]
+    non_nan_entries = features.index  # take all entries here
```
Cryaaa, line 403: Hey @jo-mueller, since Robert changed the code for clustering already I would merge that PR first and then implement this change you are suggesting on that branch, since I think it makes more sense to keep Roberts changes, as he is adding functionality there!

napari_clusters_plotter/_dimensionality_reduction.py
```diff
@@ -371,89 +374,150 @@ def update_properties_list(self):
...
+    https://pypi.org/project/umap-learn/
+    """
+    print("Selected labels layer: " + str(labels_layer))
+    print("Selected measurements: " + str(selected_measurements_list))
+
+    features = get_layer_tabular_data(labels_layer)
+
+    # only select the columns the user requested and drop NaNs
+    properties_to_reduce = features[selected_measurements_list]
+    non_nan_entries = features.dropna().index
```
Cryaaa, line 459: @jo-mueller I like this change to the structure of the dimensionality reduction widget!

napari_clusters_plotter/_utilities.py
```diff
@@ -32,11 +32,12 @@ def get_layer_tabular_data(layer):
     return None
 
 
-def add_column_to_layer_tabular_data(layer, column_name, data):
```
Cryaaa, line 35: @jo-mueller, if we change the arguments in the definition do we additionally need to go through all the other instances where this function is called and should I test that first before merging?

Cryaaa (changes requested):
Hey @jo-mueller,
I went throught the code and really like some of the changes and have added some comments. I do have some worries though that the functionality will be broken if we change the utility function definitions though....

napari_clusters_plotter/_utilities.py
jo-mueller: Hey @Cryaaa , thanks for the feedback!

Let's say I tested it a few times from Napari and did not run into any issues. But maybe it would make sense to not change the way `add_tabular_data` works but rather creating a new function for this purpose and keep the current function for backwards compatibility - just in case.

Another option: Create the to be added column of data )along with the `NaN`s in the correct position and then write this data to the tabular data with the existing function.

napari_clusters_plotter/_clustering.py
jo-mueller: Hey @Cryaaa , agree, makes a lot of sense. Let's revisit this PR later when merging the other one has been finished, I think we will run into some merge conflicts anyway.

PR #66, opened 2022-03-26
Add clustering methods

haesleinhuepf:
Hi Ryan @Cryaaa ,

this adds three new clustering methods to the clusters-plotter:
* [Gaussian Mixture Model (GMM)](https://scikit-learn.org/stable/modules/mixture.html)
* [Mean Shift (MS)](https://scikit-learn.org/stable/auto_examples/cluster/plot_mean_shift.html#sphx-glr-auto-examples-cluster-plot-mean-shift-py)
* [Agglomerative clustering (AC)](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html)

I also changed some of the clustering code to
* remove code duplication in the context of standard scaling
* remoded "SCALING" value from the "CLUSTER_ID" column name. I found that confusing. Happy to bring it back if it was important.

I wrote tests and tested the methods manually via the graphical user interface.

Would you mind reviewing the code @Cryaaa ? I don't want to bother Laura @lazigu until she's back :-)

No hurry btw. enjoy your weekend ☀️ 

Thanks!
Robert

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,10 +60,6 @@ visualization in the image, turn off the visibility of the analysed labels layer
 
 ![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/plot_interactive.png)
 
-Hold down the SHIFT key while annotating regions in the plot to manually select multiple clusters.
-
-![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/multi-select-manual-clustering.gif)
-
 You can also select a labeled object in the original labels layer (not "cluster_ids_in_space" layer) using the `Pick`
 mode in napari and see which data point in the plot it corresponds to.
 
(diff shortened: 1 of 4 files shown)


Cryaaa (commented):
Hey Robert @haesleinhuepf,
Again I really like the changes that have been made, it keeps the code more readable for sure. I tried the additional algorithms on some tribolium data and found a Warning that pops up with mean shift clustering if you set the umber of samples low (I had it at 5):

C:\Users\ryans\anaconda3\envs\ncp_dev_v2\lib\site-packages\sklearn\cluster\_mean_shift.py:246: UserWarning: Binning data failed with provided bin_size=0.000000, using data points as seeds.!

I don't actually know anything about the clustering method but should we just assume that the value might have been to low for this specific dataset?

PR #67, opened 2022-03-26
Multiple selections in different colours in the scatter plot

haesleinhuepf:
Hi Marcelo @zoccoler ,

this PR updates the manual selection code in the clusters-plotter so that users can select multiple different regions/clusters when holding SHIFT.

multi-select-manual-clustering

Would you mind reviewing the code? I don't want to bother Laura @lazigu until she's back :-)

No hurry btw. enjoy your weekend ☀️

Thanks!
Robert

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ visualization in the image, turn off the visibility of the analysed labels layer
 
 ![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/plot_interactive.png)
 
+Hold down the SHIFT key while annotating regions in the plot to manually select multiple clusters.
+
+![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/multi-select-manual-clustering.gif)
+
 You can also select a labeled object in the original labels layer (not "cluster_ids_in_space" layer) using the `Pick`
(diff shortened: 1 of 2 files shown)


Cryaaa (commented):
So I have tried out the branch and it is working but I get a warning everytime i try and cluster with shift which reads: 
See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy!
c:\users\ryans\onedrive\documents\bia pol stuff\napari plugin development\napari cluster plugin new\napari-clusters-plotter\napari_clusters_plotter\_plotter.py:247: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame

It could be avoided if we delete the cluster ID column and re-insert it basically with the new value from what I have gathered. On the other hand we could ignoew the error? I think the warning is thrown due to some indexing mix up that could happen when assigning new values...

zoccoler (approved):

Cryaaa:
Hey Marcelo @zoccoler !
Thanks for the quick changes you made to the code! I will just give it a try and then merge it if its fine on my end too

PR #70, opened 2022-03-31
Fix nans in measurements2

jo-mueller:
Closes #65 

Another approach to handling NaNs in the measured properties. I created a decorator for all functions that need to handle data which may potentially be polluted with NaNs.

The decorator assumes that the data is 
* passed as an argument
* the first argument

I think these are reasonable assumptions. The NaNs are then cleared from the data, fed to the decorated function and the original dimensions are restored before returning the data, see [here](https://github.com/BiAPoL/napari-clusters-plotter/blob/760918e2b823f4ef2a25e2ff24e9264f7c20baa6/napari_clusters_plotter/_utilities.py#L45).

Let me know what you think!

### DIFF
--- a/napari_clusters_plotter/_clustering.py
+++ b/napari_clusters_plotter/_clustering.py
@@ -19,6 +19,7 @@ from qtpy.QtWidgets import (
 
 from ._utilities import (
     add_column_to_layer_tabular_data,
+    catch_NaNs,
     get_layer_tabular_data,
     restore_defaults,
     widgets_inactive,
@@ -522,6 +523,7 @@ class ClusteringWidget(QWidget):
         show_table(self.viewer, labels_layer)
 
 
+@catch_NaNs
 def mean_shift(measurements, quantile=0.2, n_samples=50):
     from sklearn.cluster import MeanShift, estimate_bandwidth
 
@@ -531,6 +533,7 @@ def mean_shift(measurements, quantile=0.2, n_samples=50):
     return ms.fit_predict(measurements)
 
 
+@catch_NaNs
 def gaussian_mixture_model(measurements, cluster_number):
     from sklearn import mixture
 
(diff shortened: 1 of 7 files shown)


Cryaaa:
Hey @jo-mueller !
I absolutely LOVE this way of handling it, keeps it really nice and tidy and also easy to implement in future! I'll just do some tests on my end and see if this works for me aswell. We might need to take a look at the tests though as there seem to be some issues with failing tests..

jo-mueller:
Hey @Cryaaa ,

thanks!

> there seem to be some issues with failing tests..

PyOpencl error in the github actions...  -_-

Cryaaa:
@jo-mueller ok so I haven't dealt much with these errors up until now... Is it just a bug we have to accept or what can we do to fix this?

jo-mueller:
Let's see if https://github.com/BiAPoL/napari-clusters-plotter/pull/70/commits/4118888e36210ae052f5789c78ffb65fe2e099ae fixes it. I basically replaced the built-in measurement function with the one from skimage and replaced a few values with NaN.

jo-mueller:
Tests failing only for macOS is super-weird.

jo-mueller:
The strangest part is that, according to the error message, an `\n` character appears in arrays that are compared:

```
 assert np.array_equal(result[~np.isnan(result)], 1 - true_class[~np.isnan(result)])
E       assert False
E        +  where False = <function array_equal at 0x10b8765e0>(array([1., 1., 0., 0., 0., 1., 1., 1., 0., 1., 0., 0., 1., 1., 0., 0., 0.,\n       1., 1.]), (1 - array([1, 1, 0, 0, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 0, 0, 0, 1, 1])))
```

Cryaaa:
@jo-mueller ooooof.. Man This is really annoying to see and (at least for me) I cannot test any mac stuff. Maybe Robert @haesleinhuepf has seen some of these (seemingly) randomly failing test results before and has some hints?

jo-mueller:
Ok now it's even more mysterious as the test fails for Python 3.8 but works for Python 3.9 ^^"

lazigu:
> Ok now it's even more mysterious as the test fails for Python 3.8 but works for Python 3.9 ^^"

Such weird things were happening to me too with some other PR 😅 And sometimes just re-running failed tests helped..

Issue #76, opened 2022-04-12
set matplotlib backend locally?


kevinyamauchi:
Hello! Thanks for the great library. I am trying to use `napari-clusters-plotter` as the plotting library for a measurement library. I've noticed that after plotting data using `napari-clusters-plotter` in a napari viewer launched from a jupyter notebook, the matplotlib plots executed in the notebook are no longer plotted in the notebook (i.e., "inline") and are instead plotted in qt widgets. I am not a matplotlib expert, but I believe it is because the backend is being set globally (see line below). Is it possible to set the backend more locally (i.e., just for plots generated by this library)?

https://github.com/BiAPoL/napari-clusters-plotter/blob/d8af60a73a90f8b1383b8aba2f39b4e8895d532b/napari_clusters_plotter/_plotter.py#L36

To reproduce:
1. launch napari from a jupyter notebook
2. open the `napari-clusters-plotter` plugin and plot some features
3. go back to the notebook and create a plot with matplotlib
4. The resulting plot is displayed in a "pop up" qt widget

jo-mueller:
Hey @kevinyamauchi @Cryaaa ,

this is a good point! The plotting backend is indeed set globally [here](https://github.com/BiAPoL/napari-clusters-plotter/blob/4e17608fbfe82c152c0c638aaf0ca89335ca22ee/napari_clusters_plotter/_plotter.py#L36). Would changing this to `inline` directly put the plot into a noteboook? If so, we could surely declare the backend within the scope of the Plotting Widget.

Anothher option I could think of, would be to make the plotting functions and the widget independent from each other. Currently, the plotting functions are implemented as 

```Python
class Plotter():
  def plot_something(self, data):
```

We could re-arrange this to 

```Python
class Plotter():
  # create canvas and stuff here

def plot_something(ax, data):
  # do the plotting
```

When the plotting functions are called from, i.e., other plugins, the `ax` argument could point to a created plot widget - but also to a matplotlib figure instance in a notebook. Would this work? I guess the backend would still have to be selected anyway 🤔

zoccoler:
Hi @kevinyamauchi @Cryaaa @jo-mueller,

From my understanding, when you open a napari viewer from a jupyter notebook, the backend is changed to qt at that point, so I am not sure that the line mentioned would be what is causing the problem. You could check the current backend with `matplotlib.get_backend()`. In fact, I commented that line out and this still happens.

[...]

PR #78, opened 2022-04-14
Removing the setting of the backend

Cryaaa:
This was causing some issues when napari was run from a notebook (see #76). My testing didn't show  any loss of functionality when we do not set the backend explicitly. If tests pass this will be merged.


Cryaaa:
@jo-mueller, @zoccoler 
Here I fix the issue with setting the backend but I can't figure out why pre-commit tests are failing. My thinking is that the merge should be fine anyway or do you guys think something needs to be changed beforehand?

PR #82, opened 2022-04-30
changed input types, added support to add result to viewer

haesleinhuepf:
Hey Ryan @Cryaaa ,

it might not be necessary to change anything in the napari-tools-menu. I _think_ with this change the function does what it's supposed to do.

image

Best,
Robert

### DIFF
--- a/napari_clusters_plotter/_annotation_to_clusters.py
+++ b/napari_clusters_plotter/_annotation_to_clusters.py
@@ -1,31 +1,34 @@
-from ._utilities import add_column_to_layer_tabular_data
-from skimage.measure import regionprops_table
-from napari.types import LabelsData
 from napari import Viewer
+from napari.types import LabelsData
 from napari_tools_menu import register_function
+from skimage.measure import regionprops_table
+
+from ._utilities import add_column_to_layer_tabular_data
 
 
-@register_function(
-    menu="Measurement > Convert annotation to cluster ID (ncp)"
-)
-def annotation_to_cluster_id(label_image: LabelsData, annotation: LabelsData, viewer: Viewer = None) -> None:
-    regionproperties = regionprops_table(label_image, intensity_image=annotation, properties=('label', 'intensity_max'))
-    data = regionproperties['intensity_max']
+@register_function(menu="Measurement > Convert annotation to cluster ID (ncp)")
+def annotation_to_cluster_id(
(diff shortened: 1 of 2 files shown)


Cryaaa:
@haesleinhuepf, cool! Thanks for looking at it I was trying to figure out how to turn off auto call and soon realised that it wasn't a problem with the tools menu at all. Does having the input annotation set to LabelsData turn that off somehow beause napari expects you to select an input layer? I also wondered why you removed the button: just simplify the whole widget? I added it because in some cases (like my embryo and serosa annotation during the master thesis) unannotated cells don't really represent a valid cluster hence setting their cluster ID to noise (-1).

haesleinhuepf:
> Does having the input annotation set to LabelsData turn that off somehow beause napari expects you to select an input layer?

Not that I'm aware of. I just use the same pattern also in other places, because `LabelsData` is a numpy-array. Thus, you could call this function also on an image, outside napari. It will then return the array of annotations. It makes it more reusable. Furthermore, it circumvents [this issue](https://github.com/haesleinhuepf/napari-workflows/issues/2).

> I also wondered why you removed the button: just simplify the whole widget? I added it because in some cases (like my embryo and serosa annotation during the master thesis) unannotated cells don't really represent a valid cluster hence setting their cluster ID to noise (-1).

Ah ja, sorry I should have explained this in the PR. We do have the convention that not-annotated objects have value 0 in [apoc (see discussiono here)](https://github.com/haesleinhuepf/apoc/pull/21) and also in the clusters plotter. Thus, I was thinking it might make sense to not introduce a new standard here.

Cryaaa:
> Ah ja, sorry I should have explained this in the PR. We do have the convention that not-annotated objects have value 0 in [apoc (see discussiono here)](https://github.com/haesleinhuepf/apoc/pull/21) and also in the clusters plotter. Thus, I was thinking it might make sense to not introduce a new standard here.

This is not really true though since unclustered objects in the cluster plotter have cluster ID -1 (this convention was introduced by sklearn for DBSCAN and most likely taken over by HDBSCAN). I added this option because I think both ways of treating the unclustered / unannotated objects makes sense in different situations. Since this is not yet at a PR level maybe we can discuss this further in person or via zoom!

haesleinhuepf:
I see! Yes, let's discuss

PR #99, opened 2022-05-12
Fix the license file

haesleinhuepf:
Hi all @lazigu, @Cryaaa @zoccoler @jo-mueller 

this is just a minor fix in the license: year first, names in the same line after. This file is parsed be programs and must be in the right format. After this change, I would expect it doesn't look like this anymore:

image

but like this:

image

I was also wondering if Johannes wants to be in the list. If so, Johannes, please use the suggestion tool of github and add yourself as part of this PR. If anybody else could merge afterwards, that would  be great.

Thanks!
Robert

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2022, DFG Cluster of Excellence "Physics of Life" TU Dresden, Robert Haase, Laura Zigutyte, Marcello Zoccoler, Ryan Savill
+Copyright (c) 2022, DFG Cluster of Excellence "Physics of Life" TU Dresden, Robert Haase, Laura Zigutyte, Marcello Zoccoler, Ryan Savill, Johannes Müller
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,4 @@
-
-Copyright (c) DFG Cluster of Excellence "Physics of Life", TU Dresden, 2021, Robert Haase, Laura Zigutyte,
-Marcello Zoccoler, Ryan Savill
+Copyright (c) 2022, DFG Cluster of Excellence "Physics of Life" TU Dresden, Robert Haase, Laura Zigutyte, Marcello Zoccoler, Ryan Savill
```
jo-mueller, line 1:
```suggestion
Copyright (c) 2022, DFG Cluster of Excellence "Physics of Life" TU Dresden, Robert Haase, Laura Zigutyte, Marcello Zoccoler, Ryan Savill, Johannes Müller
```

jo-mueller (changes requested):

jo-mueller (approved):

PR #100, opened 2022-05-13
Comparing cluster visualisation methods and reverting to dictionary mapping

### LINKED ISSUE #97: Test best visualisation methods
In order to find out which method is fastest for visualising cluster identities we should have some testing notebooks which compare 

- pyclesperanto cluster image generation
- numpy cluster image generation (possibly numba accelerated? 
- use new colormap for cluster visualisation (for timelapse connect timepoint changed event to selection of new colormap)

Should be done before #81 since we could then decide which method to implement and if it makes sense to keep pyclesperanto at all

Cryaaa:
@haesleinhuepf, @jo-mueller and @zoccoler: In this PR there are the notebooks I created for benchmarking the different cluster visualisation methods. I have also changed the code to move back to the visualisation using dictionary colormaps as this was the overall fastest (non-pyclesperanto) version of creating a cluster image. I will add a function to export the actual cluster image in a separate PR. If you have any thoughts I'm happy to discuss them here :)

would close #97


haesleinhuepf:
Hi Ryan @Cryaaa ,

I just to find my way through the code and it's a bit too much for my tasted. Would you mind setting up two things:
* [ ] A python file with all kinds of utility functions; e.g. `generate_cluster_image_cle`, `generate_cluster_image`, ... and `generate_label_to_cluster_color_mapping` appear copy&pasted between the notebooks. Goal is that the actual test-notebooks are clean and clear.
* [ ] A notebook that generates the test data you are using. I cannot run/reproduce your notebooks because I don't have the data. Consider using synthetic data such as a blobs label image like shown [here](https://github.com/clEsperanto/pyclesperanto_prototype/blob/master/benchmarks/connected_component_labeling.ipynb)
* [ ] A notebook that runs the tests and plots the results.
* [ ] Also consider running the tests/time measurements multiple times to exclude warmup effects.

All that has low priority from my perspective. I would like to not change the clusters-plotter before we have a clear picture of where the bottleneck actually is.

Thanks for working on this!

Cryaaa:
@haesleinhuepf,
Thanks for finding out why it takes so long, that makes sense though and I agree that it is better to work with cluster images in that case. So no worries!

PR #105, opened 2022-05-19
make option to have tracking data to plot

Cryaaa:
For example Compatibility with Ko's dataset restored

### DIFF
--- a/napari_clusters_plotter/_plotter.py
+++ b/napari_clusters_plotter/_plotter.py
@@ -264,7 +264,6 @@ class PlotterWidget(QWidget):
                 self.plot_x_axis_name,
                 self.plot_y_axis_name,
                 plot_cluster_name=clustering_ID,
-                tracking_data=self.tracking_checkbox.checkState(),
             )
 
         # Canvas Widget that displays the 'figure', it takes the 'figure' instance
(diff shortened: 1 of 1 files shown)


Cryaaa:
Does anybody know why it can't import the tests anymore? the files I changed are minimal.....

zoccoler:
Hi Ryan, 

Could it be because we exhausted the [time limit for tests](https://github.com/organizations/BiAPoL/settings/billing) of the BiA-PoL account for this month?

napari_clusters_plotter/_plotter.py
```diff
@@ -333,6 +335,12 @@ def manual_clustering_method(inside):
         update_button = QPushButton("Update Axes/Clustering Selection Boxes")
         update_container.layout().addWidget(update_button)
 
+        # Update measurements button
```
lazigu, line 338:
```suggestion
        # Tracking data checkbox
```

haesleinhuepf:
I'm curious: Would it also be possible to avoid having that checkbox and instead checking if the `layer.features` do contain a `frame` column or not? I'm asking because if a timelapse dataset is loaded (`len(viewer.dims.current_step) == 4`) an no column 'frame' is present, we could visualize Ko's data, correct?

Cryaaa:
That is true and we could do it that way. On the other hand I’d the column is missing no error will pop up and the cluster image would be wrong. I do like the idea of removing checkboxes though and this was just a quick fix for the deep lineage meeting…

haesleinhuepf:
We could make a warning pop up 🌞

Cryaaa:
Actually, I think a warning everytime the plot is redrawn is annoying. I think we'll do the automatic detection and then make it very clear in the documentation.

PR #109, opened 2022-06-06
fix tests

haesleinhuepf:
I'm attempting to fix tests in this PR. I'll give a go for reviewing once I managed. 

I'm basically following up the discussion here:
https://github.com/inducer/pyopencl/issues/567

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -68,11 +68,10 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          conda install -c defaults scipy
-          conda install -y pyopencl hdbscan
-          python -m pip install --upgrade pip
-
-          pip install pytest pytest-cov PyQt5 pytest-qt pytest-xvfb
+          python --version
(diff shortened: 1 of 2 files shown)


Cryaaa:
So we already had the tests that take ages pop up here and there. @lazigu and I usually cancel them and re-run them in the case that they take ages and we both have no idea what causes this. Sometimes a PR which only changes the reader or a similar minor change fixes this but I’m guessing it’s a difficult to debug problem 😬

lazigu:
Hi Robert @haesleinhuepf,
Yes weirdly as @Cryaaa said cancelling tests and re-running failed ones helps to get them passing and quite fast instead of buffering for ages. This is quite an old problem, no idea why but I was happy with such a "workaround". And for me this was a case not only with tests on Mac but sometimes Linux or Windows, just depends which check was running the last 😕

haesleinhuepf:
No worries @lazigu  :-)

From my perspective, github-CI and automated testing are also research projects. And we're _not_ working in this research field ;-)

I suggest we merge this (If nobody disagrees) and can then update the branch in #108 and run the tests there as well.

PR #111, opened 2022-06-07
update __version__

haesleinhuepf:
Hi Ryan @Cryaaa ,

just a minor note: Whenever you update the version in setup.cfg, e.g. before releasing, you also need to do it in this file ;-)

Thanks!

Robert


Cryaaa:
ahhhhh thanks for the hint. As you probably have seen, I haven't done that in a while :D

PR #112, opened 2022-06-07
refactor colormap and put it in the tools menu

haesleinhuepf:
Hi Ryan @Cryaaa ,

I would like to make the clusters-plotter colormap accessible from the menu so that I can apply it to any label image. I found this solution by refactoring the pre-existing code. I'm just not sure if the `num_labels` parameter of the function `get_nice_color_map_rgba_dict` is really necessary or if it's possible to avoid it. Happy to chat about it in the coming days.

Thanks!
Robert

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ A table with the measurements will open:
 
 Afterwards, you can save and/or close the measurement table. Also, close the Measure widget.
 If you are uploading your own measurements make sure that there is a column that specifies the which measurement belongs to which label
-by adding a column with the name "label". If you don't specify this column it will be assumed that measurements start at 1 and each 
+by adding a column with the name "label". If you don't specify this column it will be assumed that measurements start at 1 and each
 column describes the next label.
 
 #### Time-Lapse Measurements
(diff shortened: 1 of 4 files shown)


Cryaaa:
@haesleinhuepf,
I changed the code a bit so that we get rid of the number if labels. This means the colour map is “only” 256 different Color’s but I guess that should be fine? If you agree I’ll merge this 😊

haesleinhuepf:
Hi Ryan @Cryaaa ,

it doesn't work anymore with the new change. The colours are different now.

image

I'm not sure if I understand why though. Your code looks good to me.

Cryaaa:
that is very weird. I had that problem yesterday on my desktop PC but not on my laptop today.... We'll have to investigate before we merge then.....

Cryaaa:
For my own sanity: I have traced down the issue to the length of the colormap it seems that once it reaches about 200 it displays the wrong colors (although the right colors are shown in the little icon). I'll have to investigate what it is exactly but it seems like a bug to me

haesleinhuepf:
Consider talking to the napari developers 😉

Cryaaa:
So this bug seems to appear for several people and it is being investigated by the napari developers currently: https://github.com/napari/napari/issues/4384  Until it's fixed on the napari side I would just stick to the way you implemented it at the beginning @haesleinhuepf .

haesleinhuepf:
See also: https://github.com/napari/napari/issues/4942

lazigu:
@haesleinhuepf, could you tell me if this PR is still relevant? I'm not sure whether I should solve the conflicts and consider merging it or close it due to it being a bit outdated...

PR #117, opened 2022-07-20
Un spaghettifying Qt code

### LINKED ISSUE #114: Selection is reset in measurements-list
In the clustering widget (and potentially also other places), the selection in the measurements-list is reset and that's hard to see for the user. 

To reproduce:
* Load data (e.g. blobs), segment it, appyly measurements.
* Use the dimensionality reduction widget, e.g. UMAP
* Use the clustering widget:
  * Select UMAP0 and UMAP1 only
  * Cluster, e.g. K-Means

After clustering, all measurments will be selected. If the user wants to also use HDBSCAN, they first have to select UMAP0 and UMAP1 again - and that's not obvious.

Potential solution: When updating the measurements-list, keep the selection.

Cryaaa:
This PR is here to try and clean the cluster plotter code a bit. Basically I have tried to put all GUI generating code into functions, which circumvents the code being copied and pasted in multiple places. I did not do this for every GUI element but focussed on the ones which I found in multiple files or found to be used multiple times in one file. For me this has not had any effect on the functionality but it would be good to see if you guys can find errors that I might have made. 

Furthermore, I have made a change to the measurement selections widget which was suggested in #114 by @haesleinhuepf . Now when the measurements are updated but there was a previous selection, it will keep the selection. So this PR would close #114.

For @lazigu: obviously there is no rush for you to look at this, I just wanted to tag you so you know about it :)


haesleinhuepf:
Wow @Cryaaa that's a huge PR. I think the only way for merging this, is to test it manually properly by 1-2 people. And I think we should do it. It nicely unspaghettifies the code :-)

lazigu:
Thanks Ryan @Cryaaa! This does make the code a lot cleaner. I will try to test it as I go around using the plugin now for my stuff.
It would be also great if the PR did not decrease coverage, which is already very low :sweat_smile: but I guess it could be tackled later

Cryaaa:
@lazigu, 
Thanks! I tried to increase coverage but most of the missed lines are in the plotter widget (we need to find a way to test manual selection) or in the measure widget (which we can’t test because of pyopencl…

lazigu:
@Cryaaa yep that questions is still hanging in the air. I was just wondering why this PR have failing tests due to decreased coverage while it mainly is just the refactoring of the existing code, so maybe it is still possible to catch some lines 🤔

Cryaaa:
@lazigu 
That might be possible but I think that would be going far beyond the point of deminishing returns. The coverage only decreases by 0.44% and I think writing those extra tests is not really worth it. If we instead remove measure functions, which need pyopencl we could increase coverage by a lot since the measure.py file is the one with the least coverage. 

I definitively don’t mind removing the measurement functions as everyone I have heard from who uses this plug-in has their own way of measuring their data..

We should probably discuss this in person though, I’ll get back to you once I am off of holiday and you are done with the thesis stress.

lazigu:
Good, in the mean time I will test if there are no bugs in this PR
Have a great holiday! :)

haesleinhuepf:
Last time I tested it ( #139 is based on this branch), the unspaghettified GUI worked well. `Go` from my side.

Great work again btw! Thanks for cooking un-spaghetti.

PR #124, opened 2022-08-05
Fixing scale inheritance

### LINKED ISSUE #120: Scale mismatch between label layer and highlighted/selected cluster
I am generating labels at sub resolution and overlaying them to the full resolution image by using the layer's scale argument.

viewer.add_labels(labelData, name='Fibers', scale=(8,8)).

This scale is not inherited by the "cluster_ids_in_space" layer generated after selecting points on the scatter plot.

Cryaaa:
This small PR would fix #120 by simply taking over the same scale parameter, as the analysed layer. Should not mess up much since the fix is very small


lazigu:
@Cryaaa Sorry for such a late reply! I tested this now, and it worked perfectly, so I will merge :)

PR #127, opened 2022-08-19
remove constraint for python version < 3.10

### LINKED ISSUE #126: Python 3.10 compatibility
Hi folks!

Quick question: what's the reason behind this line?

https://github.com/BiAPoL/napari-clusters-plotter/blob/815ea7036758d22db085ae18869e3f9804943e63/setup.cfg#L34

We are running into this p[y310 migration issue for another package that depends on this project](https://github.com/conda-forge/morphometrics-feedstock/pull/1), where it cannot be built due to the restrictive Python pinning.

Thanks!

haesleinhuepf:
Hi Laura @lazigu ,

do you see any issue with removing the constraint python < 3.10? 

I just tested UMAP in the clusters-plotter like this and it worked like a charm:
```
conda create --name test2 python==3.10 napari napari-segment-blobs-and-things-with-membranes -c conda-forge
conda activate test2
git clone https://github.com/BiAPoL/napari-clusters-plotter.git
cd napari-clusters-plotter
notepad setup.py
```
... remove constraint `, < 3.10`

```
pip install -e .
```
Then I tested the ncp/measurement and dimensionality reduction UMAP on blobs.tif and a corresponding label image.

Let me know what you think! If it's ok to merge it (I'm happy to do the merge and release) this closes #126 

Best,
Robert

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,7 @@ classifiers =
 
 [options]
 packages = find:
-python_requires = >=3.7, <3.10
+python_requires = >=3.7
 
 # add your package requirements here
 install_requires =


lazigu:
Hi Robert @haesleinhuepf,

I tested it also with my data, and indeed, UMAP works perfectly, so the issue for python 3.10 was resolved, and we can remove the pinned version.
Let me just figure out why this change causes failing tests. I can do the release, it's fine :) I would also review one small PR from Ryan at the same time to have it in the next release

Cryaaa:
So I think we should be able to remove this constraint, that’s always just due to numba which is updated later than numpy usually. What is alarming though is that the tests fail, haven’t looked into it yet though.

lazigu:
Apparently failed t-sne test comes because of this [commit](https://github.com/scikit-learn/scikit-learn/commit/b54a5289d5090346093ec36dc117ec2a72420563) in scikit-learn. I will make a PR to fix this failing test

Issue #81, opened 2022-04-29
Remove pyclesperanto prototype from dependencies


Cryaaa:
- [ ] We could generate the cluster image with numpy IF pyclesperanto is not installed (it might also be possible to do the dictionary colormapping visualisation on a per timepoint basis). 
- [x] Furthermore we should move measurement out of the plugin in general.

haesleinhuepf:
The measurements are already available in [napari-pyclesperanto-assistant](). I was just adding timelapse-support there. A release is imminent.

![image](https://user-images.githubusercontent.com/12660498/166109500-16b9030e-913e-44b1-849f-77b0765d46fe.png)

haesleinhuepf:
> We could generate the cluster image with numpy IF pyclesperanto is not installed

This function may be helpful:
https://github.com/haesleinhuepf/napari-skimage-regionprops/blob/master/napari_skimage_regionprops/_parametric_images.py#L5

It uses pyclesperanto as optional dependency and numpy otherwise.

Cryaaa:
@haesleinhuepf, awesome! I just have to take a quick look at our functions and see how directly we can implement this version instead!

haesleinhuepf:
Btw. the measurements are already available in napari-pyclesperanto-assistant:
https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/ad72ecae7d017ba63f579f05dd00be627f36e89b/napari_pyclesperanto_assistant/_advanced_statistics.py#L9

Thus, we can safely remove the measurements from the clusters plotter.

The only thing that's still there is replacing this section:
https://github.com/BiAPoL/napari-clusters-plotter/blob/a9754a310b790367d42df2727de050a5cfcec0a6/napari_clusters_plotter/_utilities.py#L125

By reusing this function:
* `napari_skimage_regionprops.relabel(labels, measurements)`
https://github.com/haesleinhuepf/napari-skimage-regionprops/blob/master/napari_skimage_regionprops/_parametric_images.py#L23

haesleinhuepf:
Hey Ryan @Cryaaa ,

I just saw this commit: https://github.com/BiAPoL/napari-clusters-plotter/commit/a31a986449e0e77221608b264d6cb340742d1adf

Note that very similar functionality already exists here: https://github.com/haesleinhuepf/napari-skimage-regionprops/blob/master/napari_skimage_regionprops/_load_csv.py

[...]

PR #140, opened 2022-10-24
Get rid of pyclesperanto dependancy

Cryaaa:
This is the PR finally removing the measurement widget and making the cluster image generation run with numpy if pyclesperanto is not installed. I have changed the Readme and added some tests for the cluster image generation to round off this PR. I know there were some adjustments that had to be made for the testing because of pyopencl and pyclesperanto so maybe there is some code floating around that still should bechanged, hence why this is a draft. 

I think this PR is pretty much top priority before merging the rest so we can better evaluate where we have problems with code coverage and where we are hitting the point of diminishing returns.

(@jo-mueller I didn't mean to request you as reviewer, I can't seem to change that now though so no worries if you don't review)


napari_clusters_plotter/_utilities.py
```diff
@@ -93,8 +92,34 @@ def generate_cluster_image(label_image, predictionlist):
...
+    it is assumed that len(predictionlist) == max(label_image)
+
+    Parameters
+    ----------
+    label_image: ndarray or dask array
+        Label image used for cluster predictions
+    predictionlist: array
+        Array containing cluster identities for each label
+    """
+    import pyclesperanto_prototype as cle
```
haesleinhuepf, line 119: This code looks very similar to this section in napari-skimage-regionprops (nsr): https://github.com/haesleinhuepf/napari-skimage-regionprops/blob/master/napari_skimage_regionprops/_parametric_images.py#L23-L38

Do you think it would be possible to call that function from nsr instead of duplicating the code?
```
from napari_skimage_regionprops import relabel
relabel(image, predictionlist)
```

[Read more about pros and cons of code duplication](https://en.wikipedia.org/wiki/Duplicate_code)

haesleinhuepf (changes requested):
Hey Ryan @Cryaaa ,

thanks for working on this! Just one proposal about avoiding code duplication. See below.

Thanks!
Robert

napari_clusters_plotter/_utilities.py
Cryaaa: @haesleinhuepf,
Sure! I'm not sure why I didn't think of that. The holidays must have brought me back into the old code duplication habbit :D

Cryaaa:
Just so it's also clear this PR would close #81 obviously

haesleinhuepf:
Let me just write this to see if github understands it:

closes #81

haesleinhuepf (approved):
LGTM 🌞 

Thanks again @Cryaaa

README.md
```diff
@@ -50,20 +50,17 @@ in the napari plugin [napari-segment-blobs-and-things-with-membranes](https://ww
 
 ### Measurements
 The first step is deriving measurements from the labelled image and the corresponding pixels in the grey-value image.
-You can use the menu `Tools > Measurement > Measure intensity, shape and neighbor counts (ncp)` for that.
+You can use the menu `Tools > Measurement > Label statistics (clEsperanto)` which is included in the [napari-pyclesperanto-assistant](https://www.napari-hub.org/plugins/napari-pyclesperanto-assistant)
```
lazigu, line 53: Silly question, but any idea why can't I see this entry in the tools menu? I updated the assistant and restarted napari, but I can only do measurements from the UI of the assistant and not the tools menu 🤔
image

Cryaaa: @lazigu 
You need to have the pyclesperanto assistant installed not just the napari assistant. Maybe we should make it more obvious in the readme? Although we do link the repository you need to install here... We could alternatively mention the scikit regionprops as well as the pyclesperanto regionprops as the scikit regionprops come preinstalled with nsc?

lazigu: Hmm, which one do you mean? I have pyclesperanto-prototype installed if you are talking about this one. Now I updated it, but such a menu still hasn't appeared. Sorry, I might be missing something very obvious, but my brain is still in a holiday mood :D I might try a fresh environment, because now I was using an old one

Cryaaa: hahahahaha no problem at all. I mean the pyclesperanto assistant which is the new link in the readme (https://www.napari-hub.org/plugins/napari-pyclesperanto-assistant) This is basically a library for all pyclesperanto functions in napari and was the precursor to the napari assistant. Nevertheless, if you are confused, then so will be other users. I will modify the Readme slightly so just tell me what you think once I've made the changes :)

lazigu: Ahhhh right, I had just napari-assistant in that environment :D I knew it was something silly I was missing..
But yes I think maybe it would be nice to mention it clearer, and also to mention non-gpu version of deriving measurements

lazigu:
I think now we should also add "optional" in this installation part (couldn't figure out how to comment on the not edited code part):

_* Install [pyopencl](https://documen.tician.de/pyopencl/), e.g. via conda_

haesleinhuepf:
I propse to hint users towards [devbio-napari](https://www.napari-hub.org/plugins/devbio-napari), because it comes with a couple of relevant tools. E.g. if users want to visualize the feature correlation matrix, they may want to install napari-apoc.

README.md
```diff
@@ -50,26 +50,33 @@ in the napari plugin [napari-segment-blobs-and-things-with-membranes](https://ww
 
 ### Measurements
 The first step is deriving measurements from the labelled image and the corresponding pixels in the grey-value image.
-You can use the menu `Tools > Measurement > Measure intensity, shape and neighbor counts (ncp)` for that.
+Since the 0.6.0 release measurement is no longer part of this plugin and you will have to use other napari plugins to measure your data.
```
lazigu, line 53:
```suggestion
Since the 0.6.0 release measurements widget is no longer part of this plugin and you will have to use other napari plugins to measure your data.
```

README.md
```diff
@@ -167,12 +181,53 @@ conda create --name ncp-env python=3.9
...
 ```
 
+
+* Activate the new environment via conda:
+
+```
+conda activate ncp-env
+```
+
+* Install napari-pyclesperanto-assistant, eg. with pip:
```
lazigu, line 213:
```suggestion
* Install napari-pyclesperanto-assistant, e.g. with pip:
```

lazigu (approved):
I added some comments for typos, otherwise, it looks good! Thanks for this, Ryan :)

Cryaaa:
@lazigu No problem and thanks for catching all the small mistakes! I'm a bit blind to typos sometimes... As you might have gathered through the readme I think we should make the next release 0.6.0 as so many things are changing. I would like to have the unspaghettifying PR request merged before the release though!

lazigu:
@Cryaaa Yes, I agree that a new release should come after all these changes. I promise I will review the unspaghettifying PR soonish

haesleinhuepf:
That has definitely time until after next week @lazigu 😉

PR #156, opened 2022-12-13
Update readme

### LINKED ISSUE #131: Cluster IDs in space is incorrent/current frame is not highlighted for a timelapse in version 0.5.2
Hi, all!

There is a bug in version 0.5.2 (not sure about 0.5.1, I haven't tested that one). However, in 0.5.0 it works perfectly.
In the image below, on the left - version 0.5.2, and on the right - 0.5.0 with exactly the same data, reg props file and manual cluster. The current timepoint is not highlighted in the plot, and manual clustering in the generated cluster_ids_in_space is wrong.
image

It would be nice if somebody could confirm that they also encounter this bug.
I cannot dig deeper right now, so I will continue using the older version and fix this when I have more time unless somebody is faster :)

lazigu:
- Added a small documentation section to make the requirement of sequential labels clearer if one wants to avoid incorrect visualisation. Thus, closes #131.
- Added a second screencast to demonstrate timelapse handling at the beginning of the Readme, credit for the gif goes to @haesleinhuepf

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,11 @@ A plugin to use with napari for clustering objects according to their properties
 
 This [napari] plugin was generated with [Cookiecutter] using with [@napari]'s [cookiecutter-napari-plugin] template.
 
-![](https://github.com/BiAPoL/napari-clusters-plotter/raw/main/images/screencast.gif)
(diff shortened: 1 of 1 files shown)


haesleinhuepf (approved):
LGTM :-)

PR #163, opened 2023-01-10
Fix tests for support other layer types PR

lazigu:
Testing whether current fixed tests pass.
More tests will be added.


lazigu:
All checks pass except codecov's due to the decreased coverage because one test (call to function for dim red widget) had to be disabled due to a timeout error.

PR #165, opened 2023-01-15
update menu locations in docs

### LINKED ISSUE #164: moved menus - update documentation
I was just cleaning up the `Tools > Measurement` menu because it became a bit crowded. There are now two new menus `Measurement maps` and `Measurement tables`. 

image

image

Todo:
* We should update the clusters plotter documentation to hint users to `Tools > Measurement tables` instead of `Tools > Measurement` (this should be easy)

Optional todo:
* We _could_ create another menu `Tools > Measurement post-processing` and put `Dimensionality reduction (ncp)` and `Clustering (ncp)` in there. I'm not sure if this would be helpful. Thus, I leave it for discussion :-)

haesleinhuepf:
This updates the documentation as proposed in #164 . this closes #164 

We could use this branch to also move the clusters-plotter's menus (optional). Up to discussion :-)


lazigu:
Thanks, Robert, 
I also added dimensionality reduction and clustering widgets to the "Measurement post-processing" menu and updated the documentation.

I will merge this when all checks are finished

PR #170, opened 2023-02-09
Disable multithreading for UMAP

### LINKED ISSUE #169: Napari crashes when calling dimensionality reduction from notebook
I just met the issue with @marabuuu and saw it earlier on my PC as well: Napari crashes (and the Jupyter kernel dies) when opening Napari from a Jupyter notebook and then calling dimensionality reduction (UMAP) from the menu.

How to reproduce:
* Open napari from a Jupyter notebook
* Load a dataset (we used 3D+t)
* Load a segmentation
* Do measurements (We used Tools > Measurement Tables > Measurement of all Frames (simpleitk)
* Run menu Tools > Measurements > Dimensionality reduction (ncp)
* Choose UMAP and click on run. 

After a while, napari will just close and an error shows up that the Jupyter kernel has died. This kind of error hints to a segmenation fault. When doing the same thing from napari that was started from the command line, it works fine.

lazigu:
Here comes the PR to close #169.
Multithreading was disabled for UMAP and left as an optional choice due to the jupyter kernel crashing without any error when running UMAP in napari opened from the notebook.


marabuuu:
When doing Dimensionality reduction using UMAP the Kernel and Napari are still running. However, the progress bar of the dimensionality reduction is still shown in the jupyter notebook but stays at 0% which could be confusing.

lazigu:
Thanks, Mara @marabuuu! It seems that since napari and umap end up running in the same thread causing the napari window to freeze, the progress bar in the notebook cannot be updated. I agree that is confusing, so I disabled verbose for umap unless multithreading is selected. Could you please test it one more time whether no progress bar is visible?

marabuuu:
Now the dimensionality reduction using UMAP is successfully finished without showing a progress bar.

marabuuu (approved):
Now the dimensionality reduction using UMAP is successfully finished without showing a progress bar.

lazigu:
Thanks, Mara @marabuuu!

PR #171, opened 2023-02-13
Disable buttons when computation is running

### LINKED ISSUE #79: Block widget buttons when progress bar is shown
**Possible minor future enhancement:**

It would be nice if buttons (restore_defaults, update, Run) were blocked so the user couldn´t be clicking around while a process is running. Unfortunately, to have stop/pause buttons is not possible because used functions (e.g. umap) are not generators i.e. do not yield an intermediate result only the final one, therefore the only way to stop a very long process before it finishes is to close napari.

lazigu:
This PR closes #79.
Now the buttons (run, update measurements, defaults) are blocked when the computation is running in a separate thread for dimensionality reduction and clustering widgets. They are activated again if an error occurs or the process finishes successfully.


marabuuu:
Disabling buttons works but can only be tested until the finishing of the dimensionality reduction (connected to pull request #170

marabuuu (approved):
Disabling buttons works but can only be tested until the finishing of the dimensionality reduction (connected to pull request https://github.com/BiAPoL/napari-clusters-plotter/pull/170)

PR #173, opened 2023-02-14
Create codecov.yml

lazigu:
This PR adds _codecov.yml_ config file to change the default setting that makes codecov check to fail if a PR decreases the coverage by more than 0%. Now the threshold will be 5%.


lazigu:
Tests are failing because they are fixed in another PR. 
Since this adds only the coverage config file, I will merge it as it is.

Issue #191, opened 2023-03-07
Allow usage of sub samples for Measurment -> Plot 


thorstenwagner:
This issue is a sub-issue of #189 : 

It would probably  make sense, to allow the user to only use a subsample for plotting, but applying the lasso region to the complete dataset. That would probably also make the plotting much faster, which is right now very slow.  Additionally, it would reduce the overplotting.

I can do the implementation, just want to discuss this with you before starting. What do you think? What would be the way you would like to see it implemented?

lazigu:
I can imagine the selection for subsampling living under a collapsible widget "advanced options" (e.g. like in the dimensionality reduction widget when umap is selected, picture below). When the user expands it, it would contain the selection parameter _n_, specifying every _n_ th data point that will be plotted. I would put there other plotting parameters later also, together with the checkbox to consider not selected data points as noise/background instead of the second cluster.

This way the UI does not change at all for users that are not interested in these features. What do you think?

![image](https://user-images.githubusercontent.com/52177660/223428236-f71e49ec-f66b-4224-afb1-9d6c0aa28f67.png)

thorstenwagner:
@thorstenwagner @lazigu 

I think this issue can be closed. From my side, the speed with the new 2D histogram is good enough to display also large datasets.

Please close it if you think the same.

PR #193, opened 2023-03-07
Set unselected datapoints as background/noise

### LINKED ISSUE #192: Label unselected datapoints as background/noise
This issue is a sub-issue of #189 : 

As soon you circled one cluster, it interprets the remaining points as a second cluster and all get visualized, which makes it very hard to identify your selected clusters:

image

I already found a quick hack by changing the cluster label 0 to -1 making noise labels.

I will now make a proper implementation which a checkbox in the plotting plugin.

thorstenwagner:
This is an implementation for this issue:
closes #192

I followed the instructions @lazigu gave in #191 .

This how it looks when you freshly load data:
image

This is how it looks when you circled your first cluster:
image

This is how it looks when you activate the checkbox:
image

The plot is automatically redrawn when the checkbox state is changed.

I automatically make the label layer invisible when I select the first cluster. Otherwise it would look like this after clustering:
image

Issue that I see:

- Expanding the collapsible box takes ages if you have a dataset with many points (like we do). Probably because it needs to redraw the plot. Any ideas on this @lazigu ?

Best,
Thorsten

### DIFF
--- a/napari_clusters_plotter/_plotter.py
+++ b/napari_clusters_plotter/_plotter.py
@@ -16,6 +16,7 @@ from ._Qt_code import (
     MplCanvas,
     MyNavigationToolbar,
     SelectFromCollection,
+    collapsible_box,
     button,
     labels_container_and_selection,
     title,
@@ -138,6 +139,7 @@ class PlotterWidget(QWidget):
         update_container, update_button = button("Update Measurements")
 
         # checkbox background
+        self.advanced_options_container = collapsible_box("Expand for advanced options")
         def checkbox_status_changed():
             if self.cluster_ids is not None:
                 clustering_ID = "MANUAL_CLUSTER_ID"
@@ -157,13 +159,14 @@ class PlotterWidget(QWidget):
         self.plot_hide_non_selected = QCheckBox()
         self.plot_hide_non_selected.stateChanged.connect(checkbox_status_changed)
         checkbox_container.layout().addWidget(self.plot_hide_non_selected)
+        self.advanced_options_container.addWidget(checkbox_container)
 
         # adding all widgets to the layout
         self.layout().addWidget(label_container)
         self.layout().addWidget(labels_layer_selection_container)
         self.layout().addWidget(axes_container)
         self.layout().addWidget(cluster_container)
-        self.layout().addWidget(checkbox_container)
+        self.layout().addWidget(self.advanced_options_container)
         self.layout().addWidget(update_container)
         self.layout().addWidget(run_container)
         self.layout().setSpacing(0)
--- a/napari_clusters_plotter/_plotter.py
+++ b/napari_clusters_plotter/_plotter.py
@@ -8,7 +8,7 @@ from napari_tools_menu import register_dock_widget
 from qtpy import QtWidgets
 from qtpy.QtCore import Qt
 from qtpy.QtGui import QGuiApplication, QIcon
-from qtpy.QtWidgets import QComboBox, QHBoxLayout, QLabel, QVBoxLayout, QWidget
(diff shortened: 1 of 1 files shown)


lazigu:
I could imagine the checkbox should be selected before doing any plotting (before clicking run). And instead of redrawing being triggered each time the status of the checkbox changes, when the selection of data points is made with the lasso/rectangle tool, the check is made what is the status of this box, and it proceeds accordingly with the visualisation of clusters. Does it make sense?

thorstenwagner:
With my current implementation you can also select it before you do any plotting.  But in case you change the status when you have data plotted, it will update the plots, which is I think what I would expect as a user?

I can also deactivate this, but I think it makes sense from the user perspective?

lazigu:
Ah alright, I did not think that from the user's perspective it would be expected to hide those points on the fly, but it makes sense.  I was thinking only about how computationally costly it is to keep redrawing the plot. I will think a bit about this. 
Marcelo @zoccoler, maybe you have an idea regarding this?

thorstenwagner:
Hi @lazigu ,

if you would not update it after changing the checkbox, then it would not be possible to get an updated visualization without label a second cluster, right?

And I guess, you only change this setting once, and then you keep it. So probably it you only do one additional replot.

I'm a bit more worried that extending the box takes 15-20s. However, that is probably solved once we get the overplotting issue under control.

thorstenwagner:
Ah, maybe also a misunderstanding:

> Expanding the collapsible box takes ages if you have a dataset with many points (like we do). Probably because it needs to redraw the plot.

The replot is not triggered because of the checkbox, its triggered because the size of plot changes whens the box the expanded.

lazigu:
Ahh then yes I misunderstood. However, it is weird for me that expanding the collapsible box would make the size of the plot bigger. It doesn't happen on my machine. I was playing around with your testing data on my personal laptop, which is very weak and slow, and I was pleasantly surprised that I could make it work (very cool!). Shall we introduce a minimum width for the plotter widget, so it wouldn't open for you a very small widget initially? Expanding the collapsible box shouldn't change the dimensions of the plot/plotter widget..

thorstenwagner:
Sure we can! Please go ahead an put whatever minimum size looks pleased for you :-)

lazigu:
After a bit more testing I could reproduce the issue of the collapsible box being slow to open if I have a plotter widget at the top of napari instead of the right. I merged this branch with the fix-axes-labels one, which implements a scrollbar to the plotter widget, and it seems that the slow-to-expand collapsible box issue is fixed, however, not to minimize it. I couldn't find the culprit though, need to dig a bit deeper. I think this could be opened as a separate issue, but a minor one, because now at least it seems to be fast to open the box. Otherwise, I will approve this PR, and I will add minimum widget width with another small PR.

thorstenwagner:
Could your try to select histogram instead of scatter plot? For me the issue is gone when I do that.

10.03.2023 16:56:43 Laura Žigutytė ***@***.***>:

lazigu:
Good to hear! But that branch is also merged with the branch that implements a scroll bar, plus in a way fixes overplotting. Therefore, I'm still suspicious whether redrawing or something else is happening somewhere when it shouldn't and happens anytime widget resizes :) 

By the way, there is one warning from pandas I think that keeps popping up (I can paste it later), I think if it's minor, and doesn't influence functionality, we should at least suppress it not to be annoying for the user.

thorstenwagner:
Did you tested the latest version? Because in got rid of this warning with this commit: https://github.com/BiAPoL/napari-clusters-plotter/pull/196/commits/a06800885a01768cb351987b7dd2f546a4f40a82


Its just in a different branch.


10.03.2023 17:13:50 Laura Žigutytė ***@***.***>:

lazigu:
Ah no I was testing this PR separately.

lazigu (approved):

PR #194, opened 2023-03-07
Fix axes labels not displayed bug

### LINKED ISSUE #182: Plot axes labels not visible
The plot axes labels are often not visible or just half-visible until the user resizes the window. 

image

One potential workaround / solution is to put the labels within the plot like in the napari-workflow-optimizer ([corresponding code](https://github.com/haesleinhuepf/napari-workflow-optimizer/blob/9465b8d06ff672d60532f4c5f90c7934319584e7/src/napari_workflow_optimizer/gui/_plotter.py#L51-L54)):

image

Of course, there are better solutions. I would keep it simple for now as we may change the plotting backend soon anyway.

lazigu:
This PR should close #182.

- It fixes the bug that makes axes labels not visible when the plot is made for the first time before resizing the widget.
- It changes saved figure's background to white from transparent one.
- It adds a scrollbar to the plotter widget.
- It adds constrained_layout=True when creating the figure, which is a solver that adjusts axes sizes.

I would appreciate some testing on laptops with different display settings @zoccoler @haesleinhuepf 

E.g. here is how it would look like when the widget is put to the top of napari:
image

### DIFF
--- a/napari_clusters_plotter/_Qt_code.py
+++ b/napari_clusters_plotter/_Qt_code.py
@@ -418,7 +418,7 @@ class SelectFromCollection:
 
 class MplCanvas(FigureCanvas):
     def __init__(self, parent=None, width=7, height=4, manual_clustering_method=None):
-        self.fig = Figure(figsize=(width, height))
+        self.fig = Figure(figsize=(width, height), constrained_layout=True)
         self.manual_clustering_method = manual_clustering_method
 
         # changing color of axes background to napari main window color
--- a/napari_clusters_plotter/_Qt_code.py
+++ b/napari_clusters_plotter/_Qt_code.py
@@ -436,9 +436,13 @@ class MplCanvas(FigureCanvas):
         self.axes.xaxis.label.set_color("white")
         self.axes.yaxis.label.set_color("white")
 
-        # changing colors of axes labels
+        # changing colors of axes ticks
         self.axes.tick_params(axis="x", colors="white")
         self.axes.tick_params(axis="y", colors="white")
+
+        # changing colors of axes labels
+        self.axes.xaxis.label.set_color("white")
+        self.axes.yaxis.label.set_color("white")
 
         super().__init__(self.fig)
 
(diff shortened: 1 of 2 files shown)


haesleinhuepf:
Note to self: review

marabuuu (approved):
Hi Laura @lazigu,
I reviewed your changes on the Platynereis dataset (from [Ozpolat, B. et al](https://doi.org/10.5281/zenodo.1063531) licensed [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/legalcode).).
I tested on both 125% and 150% display zoom and generally the axes labels are always visible for me. Also the scrollbar works.
Only thing I encountered was that the layer list and the right side of napari window are displayed in a very squeezed and small way at 125% zoom (see screenshots):

20230313_ncp_review2
20230313_ncp_review3
But maybe this is also only happening on my display and not so important.

At 150% zoom I did not encounter this. The visualisation generally also works when draging the plotter widget to the top of the napari GUI:
20230313_ncp_review4

In general, I am approving this pull request.

Best,
Mara

lazigu:
Thanks a lot @marabuuu!
I now also added the minimum width of the widget, could you please test how it opens the widget now?
On my laptop, it opens the widget even bigger size than the set minimum (not so squished as on yours) but does not allow the widget to be smaller than the defined size. I set it now as 400 px but not sure if it's the optimal value.

marabuuu:
I tried it out @lazigu and my problem is that when I try to open the `Plotter Widget` I get the following error message:

```
File c:\users\maral\documents\github\napari-clusters-plotter\napari_clusters_plotter\_plotter.py:190, in PlotterWidget.__init__(self=<napari_clusters_plotter._plotter.PlotterWidget object>, napari_viewer=Viewer(axes=Axes(visible=False, labels=True, col...._transform_active_layer at 0x0000017805460670>}))
    188 self.layout.addWidget(axes_container, alignment=Qt.AlignTop)
    189 self.layout.addWidget(cluster_container, alignment=Qt.AlignTop)
--> 190 self.layout().addWidget(self.advanced_options_container)
        self.layout = <PyQt5.QtWidgets.QVBoxLayout object at 0x0000017828675550>
        self.advanced_options_container = <superqt.collapsible._collapsible.QCollapsible object at 0x000001781E65D9D0>
        self = <napari_clusters_plotter._plotter.PlotterWidget object at 0x00000178215AE820>
    191 self.layout.addWidget(update_container, alignment=Qt.AlignTop)
    192 self.layout.addWidget(run_container, alignment=Qt.AlignTop)

TypeError: 'QVBoxLayout' object is not callable
```

lazigu:
Whops it was branches merging bug that I missed. Now it should be fixed. Thanks, @marabuuu for catching it!

lazigu:
Marcelo, thanks a lot for changes here, they really make sense!!
Btw we have tox file but we are not running tests with tox, you would have to edit this file https://github.com/BiAPoL/napari-clusters-plotter/blob/main/.github/workflows/test_and_deploy.yml
But I thought I would just wait till PR in napari-skimage-regionprops is merged :)

zoccoler:
Hi Laura @lazigu ,

I was about to write you here to explain these small changes but you got me first 😆 
Yep, I had forgotten about this fact the we do not use tox, I was experimenting hehe 
I agree that fixing it in napari-skimage-regionprops will solve it now.

About the changes, they are mostly for some standardization purposes, I did not bring any new fixes. I have tested and everything seems to be working properly. Also, tests pass locally for me.

Also, good idea using `constrained_layout`! It seems the best way to do it. :)

All good from my side, thanks for implementing this!

PR #197, opened 2023-03-11
Support non sequential label images and unsorted measurements

haesleinhuepf:
Hi Laura @lazigu , hi Mara @marabuuu ,

similar to [recent changes](https://github.com/haesleinhuepf/napari-skimage-regionprops/pull/67) in napari-skimage-regionprops, I'm adding support for unsorted measurements and non-sequential label images here. The trick is to use `skimage.utils.map_array` and this requires to have a list of label IDs for the corresponding predictions.

A python-test was added to see if it works on normal images and on timelapse-data. I also checked that it works from the GUI, but only tested the normal-image case with blobs.gif. Would you mind testing it again, also with timelapse data, to be sure that I didn't screw anything up?

Big thanks!

Best,
Robert

### DIFF
--- a/napari_clusters_plotter/_plotter.py
+++ b/napari_clusters_plotter/_plotter.py
@@ -401,7 +401,7 @@ class PlotterWidget(QWidget):
                             for i in range(int(max_timepoint))
                         ]
                     else:
-                        label_id_lists_per_timepoint = [
+                        label_id_list_per_timepoint = [
                             features[plot_cluster_name].tolist()
                             for i in range(self.analysed_layer.data.shape[0])
                         ]
--- a/napari_clusters_plotter/_plotter.py
+++ b/napari_clusters_plotter/_plotter.py
@@ -328,6 +328,7 @@ class PlotterWidget(QWidget):
         ):
             # fill all prediction nan values with -1 -> turns them
             # into noise points
+            self.label_ids = features['label']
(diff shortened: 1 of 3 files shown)


haesleinhuepf:
The failing test (`dimensionality reduction / test_bad_measurements`) seems unrelated to my changes 👼

marabuuu:
Hi Robert @haesleinhuepf,


I investigated the timelapse compatibility on the Platynereis dataset and it seems to work all fine. 

Steps I did in Detail:
- used from `stardist.matching` `group_matching_labels` to get non sequential label images 
- opened nuclei and membrane channel as well as grouped labels in napari GUI
- measured size, intensity and shape using `Regionprops of all frames (nsr)`
- Performed DR using UMAP
- when using `Clustering Widget (ncp)` to use the GMM I get the following error:

   ModuleNotFoundError: No module named 'deprecated'

  but the GMM is still applied

- using the `Plotter Widget (ncp)` works as well
- visualisation was tested by plotting `indexv against `frame` and checking label numbers using lasso tool -> works fine

Best,
Mara

lazigu (approved):
Tests are passing and after testing with two different non-sequential time-lapse datasets we didn't encounter any bugs, so I'll merge this, and now we have one of the bigger bugs fixed, yay!

PR #203, opened 2023-03-14
Use unique elements of cluster IDs to create colormap

thorstenwagner:
Hi! 

When we create the dictionary for the colormap, it's only necessary to loop over the unique elements. This reduces the runtime (2d histogram) for my 35 million dataset by another 5 seconds.

Best,
Thorsten


lazigu:
Great! Will merge this, when we have tests fixed :)

PR #205, opened 2023-03-14
Allow flexible cluster ids

### LINKED ISSUE #204: Hide non-selected only works for manual selections
If clusters are built from clustering methods, the hiding of non-selected does not work because it is fixed to 'MAUAL' [here](https://github.com/BiAPoL/napari-clusters-plotter/blob/aed263be9f7ec5fafb11c2cc7b1a10f729938da2/napari_clusters_plotter/_plotter.py#L153).

zoccoler:
This should fix #204

### DIFF
--- a/napari_clusters_plotter/_plotter.py
+++ b/napari_clusters_plotter/_plotter.py
@@ -150,7 +150,7 @@ class PlotterWidget(QWidget):
 
         def checkbox_status_changed():
             if self.cluster_ids is not None:
-                clustering_ID = "MANUAL_CLUSTER_ID"
+                clustering_ID = self.cluster_ids.name
                 features = get_layer_tabular_data(self.analysed_layer)
 
                 # redraw the whole plot


lazigu:
Marcelo @zoccoler, I think this should be marked as ready for review and merged. The problem with algorithmic clustering is related to plotting the 2D histogram in general, and not hiding not selected clusters, which is fixed by this PR.

lazigu:
Accidentally marked this as ready for review.

It turns out, hiding non-selected clusters is not necessary for algorithmic clustering results, because that is already done by default, e.g. hdbscan detects noise points (non-selected grey points), which are not visualized:
image

Checking the "hide non-selected clusters" would hide in this case the first (orange) cluster:
image

I think for correctness we should rename the checkbox or add a tooltip, and leave it enabled only for manual clsutering results.

lazigu:
I will merge this because it is needed to fix the issue of visualizing algorithmic clustering results in the 2D histogram plot, and deal with the issue described in my comment above in another PR.

PR #214, opened 2023-03-17
Don't reset zoom on redraw

thorstenwagner:
This PR should be merged after #196 was merged.

When you want to selected multiple clusters and use the zoom tool its quite annoying that it always resets the zoom afterwards:
reset_zoom

This PR fixes that issue. The zoom is not reset after you for example draw a cluster:
no_reset_zoom

Most of the commits come from #196 - this is the only relevant  change: https://github.com/BiAPoL/napari-clusters-plotter/pull/214/commits/6b703c827b75feb6c38cec8145aaad7b2d48fd1a


lazigu:
Hi @thorstenwagner,
Just letting you know that I merged today #196, therefore, this PR could now be updated from _main_.

lazigu (approved):
LGTM, thanks for this!

Issue #215, opened 2023-03-17
UMAP not working due to Numpy version


jo-mueller:
Hi, 

I was just testing some things and noticed that the UMAP was not functional in a fresh devbio-napari installation. 

## To reproduce

I create some sample data with this code:

```Python
import napari
from skimage import data, measure, filters
from napari_clusters_plotter._dimensionality_reduction import DimensionalityReductionWidget

image = data.human_mitosis()

labels = measure.label(image > filters.threshold_otsu(image))
features = measure.regionprops_table(labels, image, properties=('label', 'area', 'eccentricity', 'mean_intensity'))

viewer = napari.Viewer()
viewer.add_image(image, name='human mitosis')
viewer.add_labels(labels, name='human mitosis labels', features=features)


widget = DimensionalityReductionWidget(viewer)
viewer.window.add_dock_widget(widget)

napari.run()
```

When you select the UMAP algoithm in the dropwdown, al the features in the features list and click on `run`, nothing happends, whereas tSNE and PCA work. When using the debugger to step through the code, I found that [this line ](https://github.com/BiAPoL/napari-clusters-plotter/blob/33dc3c05247b230ee2497b3e488f6db63a2c5741/napari_clusters_plotter/_dimensionality_reduction.py#L561)throws an error:

```Python
result = umap(
    properties_to_reduce,
    n_neigh=n_neighbours,
    n_components=n_components,
    verbose=False,
                )
```

The error comes from numba: `"SystemError: initialization of _internal failed without raising an exception"`. The error has been described [here](https://stackoverflow.com/questions/74947992/how-to-remove-the-error-systemerror-initialization-of-internal-failed-without), with a possible fix being to downgrade numpy to 1.23.5.

Doing `mamba install numpy==1.23.5` in my environment fixed the issue for me.

## Possible fix

Pinning numpy to `numpy<1.24.0´ could do the trick, although I don't like pinning numpy. he problem has been observed in [quite a few packages](https://github.com/numba/numba/issues/8615#issuecomment-1357865375), though.

ziw-liu:
Can this pin be removed now? NumPy 1.23 is not supported per [SPEC-0](https://scientific-python.org/specs/spec-0000/) and is incompatible with other packages (e.g. latest zarr).

Cryaaa:
hey @ziw-liu,
[...]

PR #219, opened 2023-03-21
Pin numpy version due to incompatibility with numba

lazigu:
See #215
The issue will be renamed, but not closed to not forget to remove the pinned version when the problem is fixed.


haesleinhuepf (approved):
LGTM

PR #221, opened 2023-03-21
Expose frame_changed method

### LINKED ISSUE #220: Standardize frame axis position for timelapse
Currently, frame is taken as the first axis of the `current_step` of the viewer.
If data with dimension > 4 is loaded, then the frame information is wrong.

We could standardize how axes order with (..., time, z, y, x) and get `current_step` as index `-4` instead of `0`.

zoccoler:
pattern = (..., t, z, y, x)

This fixes #220


zoccoler:
To do: check data shape before looking for frame in axis -4.

zoccoler:
Converting to Draft until we figure if this makes sense in #220

Issue #103, opened 2022-05-18
Use clusters-plotter on points data


jo-mueller:
It would be super nice to use the clusters plotter on points data. Implementation could be relatively straightforward as points layers also have a `points.features` attribute where measurements could live. The resulting cluster ID could be stored in an additional column, just as we already do it with labels layers.

The color of points can be changed as follows (I think it max be the same as for labels data):
```Python
features = {'values': [0, 0.5, 1, 1.5, ....]}
viewer.add_points(points_data, color_map='hot', face_color='values', features=features)
```

Cryaaa:
Implemented in https://github.com/BiAPoL/napari-clusters-plotter/pull/229

Issue #118, opened 2022-07-22
Feature Request: Use other layer types


Michael-L-Miller:
This plugin is great.

My images are very large though so using a 'label' layer is very slow and memory intense. It would be great if support was added for other layer types, such as 'points'.

For context, I perform segmentation outside of Napari and have the object centers as xy coordinates in a separate CSV file. From the CSV file, I visualize the object centers in Napari as 'points'.

Michael-L-Miller:
That’s great to hear. Thanks for the follow up. I’ll keep an eye out for the next release.

Michael-L-Miller:
I found a work around that allows me to keep the label layer fast. I generate the label array on a sub resolution (1/8th size). This revealed another possible issue which hopefully will be easy to address. I will post it in a separate ticket for tracking purposes.

PR #229, opened 2023-03-23
Support other layer types

lazigu:
Todo list:

- add documentation
- add example data (or add recommendation to use the plugin with nppas that already has example data)
- implement support for points layer


jo-mueller:
To whom it may concern @haesleinhuepf @stefanhahmann @zoccoler @thawn @marabuuu 

I finally got around to finish this project, clean up some code and fix some smaller related issues that led to tests not passing anymore. I realize that the changes on this branch  are HUGE, so that thorough code review may be difficult if not impossible. If you could test it in some use cases though, that would already add some validity to it and finally clear the way for this to be used for all sorts of layers.

Would be very happy to see this through :)

Best, Johannes

jo-mueller:
Hi @zoccoler @thawn @Cryaaa @haesleinhuepf @stefanhahmann @marabuuu , sorry for spamming - would any of you have some time to give this not a deeper look, but at least try whether it works on your end/for your usecase?

I am afraid that the clusters plotter is falling into a bit of disrepair at the moment. With more features coming in (e.g. PR #257 ) and this one having rather extensive changes, I think it would be helpful for other contributors to move their contributions forward to see this through so that others can adapt.

Thanks!

Cryaaa:
Hey @jo-mueller,
firstly thanks for taking care of this huge PR and making it work. I definitely agree that this should be merged before anything else. I can test it with some of the previous functionality checks I still have but do you maybe have some toy datasets for other layer types that I can just check on my end? Maybe you have something, which you used which is easy for me to download and try?

jo-mueller:
A few samples I could imagine to work well:

- 2D timelapse: https://github.com/BiAPoL/PoL-BioImage-Analysis-TS-Early-Career-Track/blob/main/data/CalibZAPWfixed_000154_max.tif 
For this I think you'll first have to convert it into a 2D+t stack (there should be something in the Utilities in the Tools menu for this) and then do a measurement for all timepoints (I also think there's a specific measurement function for this in napari-skimage regionprops)
- 3D single timepoint: cells3d?
- 2D single timepoint: human mitosis?

Cryaaa:
@jo-mueller 
Actually I meant some points or Surface data examples if we have any. I have quite some test cases for the original labels layer and will try those tomorrow!

stefanhahmann:
Resolves https://github.com/BiAPoL/napari-clusters-plotter/issues/103 and https://github.com/BiAPoL/napari-clusters-plotter/issues/118

- [ ] "Manual" Test Points Layer (new)
- [ ] "Manual" Test Surface Layer (new)
- [ ] "Manual" Test 4D pseudo Labels - "old" tracking data (existing)
- [ ] "Manual" Test Vectors (new)
- [ ] Test 2D Labels, e.g. test with built-in Napari stuff (Human Mytosis)

jo-mueller:
## Code snippets for test data

```
n_points = 100
points = np.random.rand(n_points, 2)
feature1 = np.random.normal(size=n_points)
feature2 = np.random.normal(size=n_points, loc=1)
annotations = 1 * (np.random.uniform(size=100) > 0.5)

viewer.add_points(points, properties={"feature1": feature1, "feature2": feature2})
```

jo-mueller:
Hi @Cryaaa ,

thanks for taking a look at this. I added a little change (8b5fa1d0698930b207c61a3538613fccb2406768) that essentially just sets the `current_frame` and `frame_id` for pointsdata in the 3D PointsData example and it gives me this:

points

jo-mueller:
And 8d476683099e7c45f49202b73db1a6edcd7ef83d does the same for 4D surface data:

surfaces

Cryaaa:
Hey @jo-mueller 

> Which I suppose it is supposed to look like 💪

Yep that looks good now!






> And [8d47668](https://github.com/BiAPoL/napari-clusters-plotter/commit/8d476683099e7c45f49202b73db1a6edcd7ef83d) does the same for 4D surface data:

Perfect!



> > For the surface layer the visualisation is strange and does not reflect the manual clusters present (will show in person during meeting)
> 
> Maybe one side note: The peculiar thing here is that the features to color the mesh are stored per vertex, but the color is shown for a triangle. So it can happen that a face of the mesh is neighbor to three vertices with a different value to visualize on each, which creates the odd color gradient on the surface above.

Its not exactly that. As you select more and more clusters with the shift button, even the clusters which you did not change (basically the orange ones) change colors in the cluster image (see video for demonstration). No idea what might be going on there though and didn't have enough brain power to investigate today :D

https://github.com/BiAPoL/napari-clusters-plotter/assets/65285466/439fd537-3394-428d-a669-3691fcea1b3d

jo-mueller:
Ouf ok that looks really weird 😅 it's probably something in the function that creates the colormap for surfaces, will investigate.

napari_clusters_plotter/_plotter.py
```diff
@@ -795,3 +791,137 @@ def run(
...
+
+        elif isinstance(self.analysed_layer, Surface):
+            cluster_data = generate_cluster_surface(
+                self.analysed_layer.data, self.cluster_ids
+            )
+
+            cluster_layer = Layer.create(
+                cluster_data,
+                {
+                    "contrast_limits": [0, self.cluster_ids.max() + 1],
```
Cryaaa, line 890: @jo-mueller 
I think the weird surface data glitch I found might be caused by the line for the contrast limits. Everytime you add a new color it would change the limits and possibly influence all colors. I don't know enough about the surface layers yet though to know why this line might be needed.....

jo-mueller: @Cryaaa Ok, that's weird. I tried to reproduce it with the example data from above and it looks perfectly fine for me 🤔

Cryaaa: @jo-mueller,
damn that is strange.... I'll have to investigate my environment then and see what might be influencing it but maybe should not try to change things too drastically

jo-mueller: Same - I'll create a fresh environment to see whether it still works alright.

Cryaaa: In another environment I have the same problem. Maybe to add some more details, I was adding new clusters each time in the video (i.e. hold shift while circling the same points each time). If I just circle the same points without holding shift it behaves as you would expect

jo-mueller: @Cryaaa ok, now I actually see what the bug is, by adding a color you mean adding a new class while holding the shift key? I'm not sure whether I would treat that as a bug rather than a feature. After all, adding a new color is supposed to add a new class color - selecting the same set of points with the shift-key pressed would only a bit of an...exotic? way of using that feature ^^

Maybe we could make the resetting the colormaps a bit more prominent with an extra button somewhere in the plot? Currently this is implicitly done by clicking somewhere in the image or making an empty selection without holding the shift-key.

Cryaaa: @jo-mueller,
let's talk this through in person later, then I can show you how the bug might be problematic!

jo-mueller: @Cryaaa could you maybe share the code snipped you used to create the problematic sample data you showed earlier?

Cryaaa: Sure

```python
import napari
import vedo
import pandas as pd
import napari_process_points_and_surfaces as nppas


sphere = vedo.shapes. IcoSphere()
features = pd.DataFrame(sphere.vertices,columns=["x","y","z"])

viewer = napari.Viewer()

layer = viewer.add_surface(nppas.to_napari_surface_data(sphere))
layer.features = features
```

jo-mueller: @Cryaaa , I think you were right about the contrast limits. Changing the contrast limits from [x, y] to [x, y+1] fixes the problem:

capture

Cryaaa: @jo-mueller Awesome, nice that it was a quick fix 🙌🏻

jo-mueller: Fixed in 9c7fb837078a910de39043d0f6ca112241b05946

Cryaaa:
@jo-mueller, @zoccoler, @stefanhahmann 
I will merge this PR if all of you agree, so that we can continue with our other changes on Thursday!

haesleinhuepf:
Great to see this merged everyone. Fantastic work! 😍

PR #237, opened 2023-03-30
Implement Histogram1D if the same measurement is selected for both axes

### LINKED ISSUE #151: Histograms
In case the user chooses the same parameter to plot as X and Y axis, we could draw a histogram instead of a scatter plot. In that case, the lasso tool should just ignore the Y-Axis and allow to select histogram-bins instead.

lazigu:
This PR implements plotting 1D histogram if the same measurement is selected for both axes. Currently, no clustering for this type of histogram is implemented. @zoccoler had an idea that clustering for such a histogram could be done with a span selector (the issue has not been created yet).
Partly closes #151, however, a separate issue regarding clustering 1D histograms should be created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,11 @@ on `Update Axes/Clustering Options` to refresh them. Click on `Plot` to draw the
 
 Currently, you can select between two types of plots: scatter plot and 2D histogram. Click on "expand for advanced
 options" to see the selection of the plot type. Clustering (manual and automatic) is possible using both types of plots.
-2D histogram is recommended if you have a very high number of data points. Under advanced options you will also find the
-checkbox determining whether not selected data points should be automatically clustered as another cluster or displayed as
-gray data points and not visualized in the generated clusters IDs layer.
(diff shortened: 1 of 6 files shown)


zoccoler:
Hi @lazigu , sorry for the delay!
I am implementing the span selector in this PR, so that it becomes complete.

zoccoler:
The histogram still needs to be painted back with the right colors. As metnioned in the commit, somthing like this may work: https://stackoverflow.com/a/49290555/11885372

This is a work in progress.

zoccoler:
Hi @lazigu, hi everyone, the span selection should be working now. The histogram colors now also get updated, having a similar behavior to the other plotting options (histogram 2d and scatter).

I renamed the `HISTOGRAM_2D` to just `HISTOGRAM` now. Whether it is 2D or not comes from the selected measurements.

If both measurements are the same, the histogram is 1D and a span selector is used instead of the lasso. Right-click should reset the clusters.

Could someone test before merging?

Best,
Marcelo

haesleinhuepf:
I love it!

image

Issue #251, opened 2023-05-04
Resetting zoom after selection of manual clustering


thorstenwagner:
Hi,

with the PR https://github.com/BiAPoL/napari-clusters-plotter/pull/214 it did not reset zoom after selecting a cluster.

With the recent release it does it again. So something broke it. Any idea what change might be responsible?

Best,
Thorsten

thorstenwagner:
Found a fix. As soon I'm allowed to open a PR, I will do it :-)

haesleinhuepf:
done in #254

PR #254, opened 2023-05-10
Fix reset zoom

thorstenwagner:
This does fix issue #251 

I'm not super happy with it, because it is a bit hacky. But it does the job.


haesleinhuepf:
I'd say it's fine. Thanks again!

Issue #269, opened 2023-09-25
Label image disappears when performing clustering on timelapse data


stefanhahmann:
Steps to reproduce:

1. Fresh installation of Napari following the instructions of https://biapol.github.io/blog/mara_lampert/getting_started_with_mambaforge_and_python/readme.html
2. Import label_image.tif contained in [label_image.zip](https://github.com/BiAPoL/napari-clusters-plotter/files/12717466/label_image.zip) (zip file contains file label_image.tif, which github did not allow to upload directly)

![label_image_import](https://github.com/BiAPoL/napari-clusters-plotter/assets/10515534/39d61e00-c5bf-43d0-ab3f-131d0c093eec)

The same label may occur in multiple frames.

3. Load measurements contained in 
[measurements.csv](https://github.com/BiAPoL/napari-clusters-plotter/files/12717476/measurements.csv) using Tools > Measurements > Load from CSV

![load_measurements](https://github.com/BiAPoL/napari-clusters-plotter/assets/10515534/3d8ae972-172c-4636-8ca8-b83b1644b78e)

The same label may occur in multiple frames. However, the combination label-frame is unique.

4. When plotting two features, e.g. Branch Sinousity and Branch displacement and making a selection, the original label image label turns black (which is unexpected) and the id layer does not really show the results.

![ncp](https://github.com/BiAPoL/napari-clusters-plotter/assets/10515534/61d7456b-ec40-4cd8-8dd6-5287fb6c0a5a)

When using the 2D-view (i.e. multiple slices per timepoint), the selection performed in the cluster plotter widget seems to be shown correct. However, the original label image is still black.

It is a bit unclear to me, if the provided data is not sufficient or if there actually is a bug in Napari Cluster Plotter

zoccoler:
Hi @stefanhahmann !

A quick thing I could notice is that you have a label `0` in the table. The `0` usually refers to the background, so I think you should avoid having that. Maybe you have to increment all labels. This may be at least part of the problem.

To better investigate that, you could check the checkbox "Show selected" on the top-left. This should allow you to check which row of the table correspond to which object, like shown in the first gif [here](https://github.com/haesleinhuepf/napari-skimage-regionprops#napari-skimage-regionprops-nsr). I am not sure if this works with time-lapse data, but I would expect it does.

stefanhahmann:
> Hi @stefanhahmann !
> 
[...]

PR #270, opened 2023-09-27
Add additional explanation regarding zero as a label

stefanhahmann:
* This was motivated by running into unexpected behaviour when using a label image and a measurement table that used `0` as a label, cf.: https://github.com/BiAPoL/napari-clusters-plotter/issues/269

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ Jump to:
 
 ### Starting point
 For clustering objects according to their properties, the starting point is a [grey-value image](example_data/blobs.tif) and a label image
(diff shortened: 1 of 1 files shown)


jo-mueller (approved):

jo-mueller:
Hi @stefanhahmann ,

I think this should be good to go, thanks for spotting and opening the issue. Where I am concerned: Feel free to merge!

Cryaaa:
Since tests are now passing after updating from main this is good to go!

PR #274, opened 2023-11-13
added explicit dependencies

### LINKED ISSUE #268: Update dependencies
Hello team!

While working on https://github.com/conda-forge/napari-clusters-plotter-feedstock/pull/13 we found:

Package listed as requirement but not imported
- joblib 

Packages not listed as requirements but directly used and imported in code
    - qtpy
    - dask-core
    - napari
    - magicgui
    - scikit-image
    - superqt
    - scipy

Could these issues be fixed for the next release?

Cheers

jo-mueller:
#268 

This adds a few dependencies to the `setup.cfg` that are, according to @goanpeca, imported in the plugin but not explicitly declared in the dependencies. Examples are, for instance, `magicgui, `scikit-image` and `scipy`, all of which are dependencies of napari and thus not explicitly declared although they should be.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ install_requires =
     napari-skimage-regionprops>=0.3.1
     hdbscan
     qtpy
-    dask-core
+    dask
     napari
     magicgui
     scikit-image
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,13 @@ install_requires =
     napari-tools-menu
     napari-skimage-regionprops>=0.3.1
     hdbscan
-    joblib
+    qtpy
+    dask-core
+    napari
+    magicgui
+    scikit-image
+    superqt
+    scipy
 
 
 [options.entry_points]


setup.cfg
```diff
@@ -47,7 +47,13 @@ install_requires =
     napari-tools-menu
     napari-skimage-regionprops>=0.3.1
     hdbscan
-    joblib
+    qtpy
```
haesleinhuepf, line 50:
```suggestion
```
I strongly discourage adding this dependency because of licensing reasons. Also we do not use any functions from this package.

haesleinhuepf: No, wait I take this back. It's the right dependency  :-)

setup.cfg
```diff
@@ -47,7 +47,13 @@ install_requires =
...
+    dask-core
```
Cryaaa, line 51: I think this is breaking the tests, maybe we can just put the whole of dask as a dependency? One problem I see is that the pypi installation for the complete dask library requires an import that looks like this: pip install "dask[complete]"  [see documentation](https://docs.dask.org/en/stable/install.html). I do not know how to specify this in requirements for conda and pip seperately though since they have differing commands. Maybe we should check how it is done i napari..

Cryaaa: In napari they seem to mention dask in the requirements like this:

    dask[array]>=2.15.0,!=2.28.0  # https://github.com/napari/napari/issues/1656
[see here](https://github.com/napari/napari/blob/6affa18d1d5782bb31532c5f6be07f4a31e9a503/setup.cfg#L50) but I think that might change at some point because their weird version pinning seems to be related to an issue with a dask update so maybe we should not specify the exact version and then the dependency solver will take care of it?

jo-mueller: @Cryaaa I think I am back to the previous error which is also failing on the main branch ^^" Thanks for spotting!

jo-mueller:
With PR #229 through, I believe this can go in as well. @Cryaaa @zoccoler ?

PR #282, opened 2023-12-07
Don't recalcuate histogram if not necessary

thorstenwagner:
Hi,

I'm making great use of this plugin, but if data gets huge (like 32 million datapoints) it becomes much slower.

For example, after the umap is loaded and I circle a cluster it takes like 17 seconds to highlight the clusters.

This is the first PR to improve this. With the proposed changes, the 2dhistogram is not recalculated if the data hasn't changed.

It already reduces the computational time from 17 seconds to like 12 seconds.

Seconds PR follows :-) 

Best,
Thorsten

### DIFF
--- a/napari_clusters_plotter/_Qt_code.py
+++ b/napari_clusters_plotter/_Qt_code.py
@@ -592,8 +592,11 @@ class MplCanvas(FigureCanvas):
         norm = None
         if log_scale:
             norm = "log"
-        if self.histogram is not None and np.array_equal(self.last_datax, data_x) and np.array_equal(self.last_datay,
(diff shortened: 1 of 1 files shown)


Cryaaa:
This looks good to me! Will
Merge once tests pass. By the way, we are restructuring the plotter code to tidy it up and improve it currently. If you have some suggestions for larger changes or more efficient approaches we should make please let us know and we can try to implement them in the next big release! Thanks for taking care of this improvement 😊

thorstenwagner:
Fantastic! Thanks for reacting so quick!

PR #283, opened 2023-12-07
Speed-up generate cluster image

thorstenwagner:
The generate_cluster_image method is the most computationally intensive part when working with large data like mine (32 million data points). As already written in #282, the total time from circling a cluster to visualizing the highlighted cluster is 17 seconds for my data.

This PR reduces it by another 6 seconds, and thus together with PR #282 from 17 seconds to 6 seconds - which is a much better user experience.

The PR works beautifully on my data, but I'm a bit skeptical that it will work with other workflows. So it would require some testing on your part :-) 

Please let me know your thoughts!

Best,
Thorsten

### DIFF
--- a/napari_clusters_plotter/_Qt_code.py
+++ b/napari_clusters_plotter/_Qt_code.py
@@ -528,6 +528,8 @@ class MplCanvas(FigureCanvas):
         self.match_napari_layout()
         self.xylim = None
         self.last_xy_labels = None
+        self.last_datax = None
+        self.last_datay = None
 
         super().__init__(self.fig)
         self.mpl_connect("draw_event", self.on_draw)
@@ -590,7 +592,14 @@ class MplCanvas(FigureCanvas):
         norm = None
         if log_scale:
             norm = "log"
-        h, xedges, yedges = np.histogram2d(data_x, data_y, bins=bin_number)
+        if self.histogram is not None and np.array_equal(self.last_datax, data_x) and np.array_equal(self.last_datay,
+                                                                                                     data_y):
+            (h, xedges, yedges) = self.histogram
+        else:
+            h, xedges, yedges = np.histogram2d(data_x, data_y, bins=bin_number)
+            self.last_datax = data_x
+            self.last_datay = data_y
+
         self.axes.imshow(
             h.T,
             extent=[xedges[0], xedges[-1], yedges[0], yedges[-1]],
--- a/napari_clusters_plotter/_utilities.py
+++ b/napari_clusters_plotter/_utilities.py
@@ -300,12 +300,10 @@ def generate_cluster_image(label_image, label_list, predictionlist):
 
     # label_list can be removed from method.
     predictionlist_new = np.array(predictionlist) + 1
-    plist = np.zeros(np.max(label_image)+1,dtype=np.uint32)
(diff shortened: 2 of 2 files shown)


thorstenwagner:
Tests are failing ^^ Lets see if I can fix it.

thorstenwagner:
Alright! Now all tests pass and its still faster :-) What do you think?

thorstenwagner:
Here are a script to evaluate the speedup:
```python
import numpy as np
from skimage.util import map_array
from time import perf_counter

shape = (1,1024,1024)
total = shape[0]*shape[1]*shape[2]
NUM_LABELS=100000
input_data = np.random.randint(NUM_LABELS,size=total).reshape(shape).astype("int64")
from_values = np.arange(NUM_LABELS)
to_values = np.copy(from_values)[::-1]

def generate_cluster_image(label_image, label_list, predictionlist):
    """
    Generates a clusters image from a label image and a list of cluster predictions,
    where each label value corresponds to the cluster identity.
    It is assumed that len(predictionlist) == max(label_image)

    Parameters
    ----------
    label_image: ndarray or dask array
        Label image used for cluster predictions
    predictionlist: Array-like
        An array containing cluster identities for each label

    Returns
    ----------
    ndarray: The clusters image as a numpy array.
    """

    # label_list can be removed from method.
    predictionlist_new = np.array(predictionlist) + 1
    plist = np.zeros(np.max(label_image) + 1, dtype=np.uint32)
    plist[label_list] = predictionlist_new

    predictionlist_new = plist

    return predictionlist_new[label_image]

def generate_cluster_image_old(label_image, label_list, predictionlist):
    """
    Generates a clusters image from a label image and a list of cluster predictions,
    where each label value corresponds to the cluster identity.
    It is assumed that len(predictionlist) == max(label_image)

    Parameters
    ----------
    label_image: ndarray or dask array
        Label image used for cluster predictions
    predictionlist: Array-like
        An array containing cluster identities for each label

    Returns
    ----------
    ndarray: The clusters image as a numpy array.
    """
    from skimage.util import map_array

    # reforming the prediction list, this is done to account
    # for cluster labels that start at 0, conveniently hdbscan
    # labelling starts at -1 for noise, removing these from the labels
    predictionlist_new = np.array(predictionlist) + 1
    label_list = np.array(label_list)

    return map_array(np.asarray(label_image), label_list, predictionlist_new).astype(
        "uint32"
    )

t1 = perf_counter()
res_new = generate_cluster_image(input, from_values, to_values)
t_new = perf_counter() - t1
print(t_new)

t1 = perf_counter()
res_old = generate_cluster_image_old(input, from_values, to_values)
t_old = perf_counter() - t1
print(t_old)

print(f"Speedup {t_old/t_new}")

print(np.array_equal(res_new,res_old))
```

The new function is like 30x-35x faster.

Cryaaa:
Hey @thorstenwagner!

> Thanks for the merge! One question: what do you want to happen if from_values and to_values are only a subset of the labels from label image?
> 
> Right now they are mapped to zero. But we can easily modify it so that they dont get changed?

I'm not sure what benefit that would have exactly? The cluster image is just supposed to show the data points that have been clustered in some way. It might also be a fringe case but if you just transfer the original label and the original label coincides with the cluster label you have misclassified objects. Other than that if you are interested in the unclassified labels you could just extract them from the table and process them separately.

PR #284, opened 2023-12-11
Convert max label to integer before using it as parameter for np.zeros

stefanhahmann:
This PR fixes a the following exception:

grafik

which was raised after this workflow:

GIF 11 12 2023 15-23-57

The consequence of the exception was that no cluster_ids_in_space layer was generated after making a freehand selection in the clusters plotter.


Cryaaa:
@stefanhahmann,
Thanks for catching this! Slightly weird that the label image is not stored as an integer array but possibly the numpy max function returns non integers by default..

PR #287, opened 2023-12-18
More efficient multiple cluster selection

thorstenwagner:
the corresponding feature column is first converted to a numpy array, then the cluster id is changed, and finally it is converted back to a numpy dataframe.

Obviously, this is not very efficient. 

This PR does it inplace, without converting to numpy, and should be more efficient.


thorstenwagner:
> @thorstenwagner, Thanks again for finding these points of improvement. LGTM and tested it on my local machine!

You're welcome :-) Happy to see this plugin thrive..

PR #288, opened 2023-12-18
Update full data only when data is changed

thorstenwagner:
In `make_2d_histogram` we we created an concatenated dataframe of all input data every time we call the method. We do that, even if the input data has not changed.

This PR change this behavior that this is only updated when input data has changed.


Cryaaa:
LGTM, also did a quick test locally to confirm functionality

PR #290, opened 2023-12-18
Update method generate_cluster_tracks()

stefanhahmann:
* In https://github.com/BiAPoL/napari-clusters-plotter/pull/286#issuecomment-1856420498 it was pointed out that

> The problem is that my method assumes that label_list only contains labels that are actually part of the label_image, which is not the case for your track data. I think it should, and the implementation of `generate_cluster_tracks` may be buggy as it produces the same label_list for each time point, regardless of the actual labels at that time point.

* This PR is dedicated to provide a solution for this:
  * Cluster tracks generates now image specific label and cluster id lists instead of repeating the complete lists containing all labels for all images
  * Generate_cluster_image() does not need to account for that fact anymore

### DIFF
--- a/napari_clusters_plotter/_utilities.py
+++ b/napari_clusters_plotter/_utilities.py
@@ -240,18 +240,20 @@ def update_properties_list(widget, exclude_list):
 
 
 def generate_cluster_tracks(analysed_layer, plot_cluster_name):
+
     features = analysed_layer.features
-    label_id_list_per_timepoint = [
-        features["label"].tolist() for i in range(analysed_layer.data.shape[0])
-    ]
-    prediction_lists_per_timepoint = [
-        features[plot_cluster_name].tolist()
-        for i in range(analysed_layer.data.shape[0])
-    ]
+    label_id_lists_per_timepoint = list()
+    prediction_lists_per_timepoint = list()
+
+    for i in range(analysed_layer.data.shape[0]):
+        labels_of_timeframe = np.unique(analysed_layer.data[i].flatten())
+        filtered_features = features[features["label"].isin(labels_of_timeframe)]
+        label_id_lists_per_timepoint.append(filtered_features["label"].tolist())
+        prediction_lists_per_timepoint.append(filtered_features[plot_cluster_name].tolist())
 
     cluster_data = dask_cluster_image_timelapse(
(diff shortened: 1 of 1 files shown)


napari_clusters_plotter/_utilities.py
```diff
@@ -327,9 +336,7 @@ def generate_cluster_image(label_image, label_list, predictionlist):
     """
 
     predictionlist_new = np.array(predictionlist) + 1
-    plist = np.zeros(
```
Cryaaa, line 330: Does reverting this change drastically change the time to execute the function? Otherwise I actually prefer a more robust function even if we don't specifically make use of this use-case anymore. If it is adding a lot of processing time then of course we can change it

stefanhahmann: There is some speedup, but no drastic speedup when reverting this.
I would still suggest to remove/revert it, since this would reduce code complexity (think of our future selves or even on newbie contributors). 
Regarding robustness: we have some more tests now that would cover non-robust changes of this in the future.

napari_clusters_plotter/_utilities.py
```diff
@@ -241,17 +241,20 @@ def update_properties_list(widget, exclude_list):
...
-    ]
-    prediction_lists_per_timepoint = [
-        features[plot_cluster_name].tolist()
-        for i in range(analysed_layer.data.shape[0])
-    ]
+    label_id_lists_per_timepoint = list()
+    prediction_lists_per_timepoint = list()
+
+    for i in range(analysed_layer.data.shape[0]):
+        labels_of_timeframe = np.unique(analysed_layer.data[i].flatten())
```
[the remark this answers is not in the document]
stefanhahmann: No. We don't actually need this. I have removed it. Thanks for spotting this.

napari_clusters_plotter/_utilities.py
Cryaaa: I guess it does reduce code complexity but in the case other people are using this function in place of the old one which uses the mapping it might be better to have a function that behaves as before. I will add some comments to the function to make clear why the extra line is there

stefanhahmann:
I think, this PR is ready to be merged.

PR #294, opened 2024-01-11
Update axes limits when plotting

thorstenwagner:
I noticed that when switching between layers, the axis limits of the plots are not updated properly. This is due to reset_zoom being called with old axis limits. This PR fixes the problem.

### DIFF
--- a/napari_clusters_plotter/_Qt_code.py
+++ b/napari_clusters_plotter/_Qt_code.py
@@ -617,6 +617,7 @@ class MplCanvas(FigureCanvas):
         )
         self.axes.set_xlim(xedges[0], xedges[-1])
         self.axes.set_ylim(yedges[0], yedges[-1])
+        self.xylim = (self.axes.get_xlim(), self.axes.get_ylim())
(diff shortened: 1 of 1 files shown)


Cryaaa:
Hey @thorstenwagner,
Is this also a problem for other plotting types like 1D Histograms and scatterplots? If so should we maybe add the line to the other functions too? Apart from that thanks for catching this and the simple fix!

thorstenwagner:
I don't have a good example for the other plotting types. Especially not one with multiple layers

thorstenwagner:
Tried it with a small dataset I've. Seems to work fine for scatter and histogram plots

thorstenwagner:
I can reproduce the issue with scatter plots

thorstenwagner:
Fixed. For histograms I can't reproduce the issue. Seems to work. Ready for merge :-)

Cryaaa:
> Fixed for scatter plots. For histograms I can't reproduce the issue. Seems to work. Ready for merge :-)

Awesome, thanks for checking this and implementing the fix! Everything LGTM

PR #296, opened 2024-01-15
Fixing the problem

### LINKED ISSUE #295: [pre-]selected layer is an image layer with no properties
Hi all,

there is a [_regression bug_](https://en.wikipedia.org/wiki/Software_regression) in the user-interface / convenience with the current version compared to 0.7.3:

If a user starts `Tools > Visualization > Plot measurements (ncp)`, they see an empty pulldown for X/Y axis, because the pre-selected layer is an image layer, that does not contain any features / measurements:

image

In 0.7.3 the window looks like this:

image

The current version makes it more complicated to use. I propose to change the behaviour back to how it was in 0.7.3: The pre-selected layer should be the currently active layer in the napari window.

## How to reproduce:

* Open blobs.tif
* Segment it using `Tools > Segmentation / Labeling > Gauss-Otsu-Labeling` (default parameters)
* Create measurements using `Tools > Measurement tables > Regionprops (skimage, nsr)`
* Run the plotter `Tools > Visualization > Plot measurements (ncp)`

Let me know if you need any further information from my side.

Best,
Robert

Cryaaa:
This PR fixes #295. I do a quick check which layers have properties or features attached to them and we select the first one in the layer list.

### DIFF
--- a/napari_clusters_plotter/_Qt_code.py
+++ b/napari_clusters_plotter/_Qt_code.py
@@ -24,6 +24,7 @@ from qtpy.QtWidgets import (
     QWidget,
 )
 from superqt import QCollapsible
+from ._utilities import check_for_properties_and_features
 
 ICON_ROOT = PathL(__file__).parent / "icons"
 MAX_WIDTH = 100
@@ -52,7 +53,7 @@ def measurements_container_and_list():
     return properties_container, properties_list
 
 
-def layer_container_and_selection():
+def layer_container_and_selection(viewer = None):
     """
     Create a container and a dropdown widget to select the layer.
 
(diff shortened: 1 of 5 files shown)


haesleinhuepf:
Thanks gor quick fix @Cryaaa ! Would it be possible to select the layer that is currently active in napari instead? That would go in line with the suggestion in #295 and is the same behaviour as in the napari-assistant for example.

Cryaaa:
> Thanks for the quick fix @Cryaaa ! Would it be possible to select the layer that is currently active in napari instead? That would go in line with the suggestion in #295 and is the same behaviour as in the napari-assistant for example.

Sure thats also a nice option and might require less code and avoid those errors popping up

napari_clusters_plotter/_Qt_code.py
```diff
@@ -67,6 +67,9 @@ def layer_container_and_selection():
     layer_select = create_widget(annotation=Layer, label="layer")
     layer_selection_container.layout().addWidget(layer_select.native)
 
+    if viewer is not None and viewer.layers.selection.active is not None:
+        layer_select.value = viewer.layers.selection.active
```
haesleinhuepf, line 71: I'm speculating this crashes when two layers are selected in napari?

Cryaaa: I tested this and at the moment when you select two layers the selection.active returns none. I’m guessing that might change some time in the future though but I’m not sure

Issue #297, opened 2024-01-18
Zoom is reset after selection of cluster


thorstenwagner:
With the PR #294 I reintroduced the problem I "fixed" in issue #254 

I try to find a solution for both problems :-/

PR #298, opened 2024-01-18
fix zooming problem

thorstenwagner:
Fix issue #297 

Now it works in both cases: The one I described in #294 and the other one described in #254


Cryaaa:
I don't see anything wrong with the code changes suggested, and since you've tested it I will merge and then test the final release version!

Issue #303, opened 2024-01-20
Decide/document if current version is backwards compatible  to 0.7.3


haesleinhuepf:
If the version on current main branch is not backwards compatible to 0.7.3, the next release should be 0.8.0 and not 0.7.4 . Also the release notes should list what users should be aware of. This should be done before #281

zoccoler:
Since there were many updates, new version was set to 0.8.0 and release notes can be found [here](https://github.com/BiAPoL/napari-clusters-plotter/releases/tag/0.8.0).

Issue #302, opened 2024-01-20
Test current version before releasing


haesleinhuepf:
I'm not sure if all use cases were properly tested. It would be cool if someone volunteers for testing them before #281 , as the tool is downloaded and used by quite some folks.

* [x] 2D labels
* [x] 3D labels
* [x] 2D + time
* [x] 3D + time
* [x] points
* [x] surface
* [x] points + time ?  (bf09d794596de073a993f5e8c942782bb1a42ff3)
* [x] surface + time ?  (9cf67049d4222163e2f9c08f4c46778eddb52729)

For the last tests, example data might be necessary (see also #301 )

Thanks!

Cryaaa:
I had made some progress on this issue but one Issue is that the [napari-skimage-regionprops](https://github.com/haesleinhuepf/napari-skimage-regionprops) function `show_table` is incompatible with surface layers currently. I have made a suggestion to fix it as a [PR](https://github.com/haesleinhuepf/napari-skimage-regionprops/pull/85) but until this gets merged and released we will need to either keep the newest cluster plotter release on hold or find a workaround for showing the table..

Cryaaa:
@zoccoler Have you tested the new features manually yet? I don't think the plugin will work for the surface layer as expected because as soon as you perform dimension reduction or clustering on the surface layer the `show_table` function will erase the surface layer data.... I don't know if you would prefer just to do a workaround which I could get started on or if we should wait until the suggested PR is merged in napari-skimage-regionprops?

zoccoler:
Hi @Cryaaa ,

I tried to fix [that PR](https://github.com/haesleinhuepf/napari-skimage-regionprops/pull/85) today, but I did not try that branch with the napari-clusters-plotter.
If you have some time, could you test if it would work?

Issue #301, opened 2024-01-20
Supported layer types in documentation


haesleinhuepf:
According to the current documentation, only label layers are supported. I recommend updating documentation before #281 

* [ ] List supoorted layer types
* [ ] Explain how to derive measurements for each layer type / list plugins which can do this
* [ ] Add at least one screenshot or screencast showing how to process a surface (because that's a cool new feature that should be highlighted)
* [ ] Add example data for new layer types so that people can try it out

Cryaaa:
Hey @zoccoler!
That would be awesome if you could take care of this. I have an example dataset which I created for testing which uses the napari process points and surfaces package which I could send to you. Maybe we could add it as an example notebook? For the points layer I also have a very basic example which I used which just converts cells to points

Cryaaa:
I have now added some example data for [surfaces](https://github.com/BiAPoL/napari-clusters-plotter/tree/main/example_data/surface_data) and [points](https://github.com/BiAPoL/napari-clusters-plotter/tree/main/example_data/points_data) which can be found using the links!

zoccoler:
Thanks!! I am working on the docs now

PR #304, opened 2024-01-22
Set up automatic deployment

### LINKED ISSUE #299: Set up automatic release on PyPi
We should update the [`test_and_deploy.yaml`](https://github.com/BiAPoL/napari-clusters-plotter/blob/main/.github/workflows/test_and_deploy.yml) to support automatic releases of the plugin through PyPi. For that, we should change the packaging inLine 115 to use 

```
python -m build .
`´`

rather than the use of bdist (`python setup.py sdist bdist_wheel`). We need an API token of the repos owner on PyPi, though.

jo-mueller:
Fixes #299 

This PR sets the correct API token for an automatic release through PyPi upon publishing a release on github. It also sets the latest versions for the involved github actions. It also uses the newer `build` functionality instead of the older `bdist wheel` etc.

I essentially copied the code from [here](https://github.com/napari/cookiecutter-napari-plugin/blob/main/%7B%7Bcookiecutter.plugin_name%7D%7D/.github/workflows/test_and_deploy.yml)

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -105,12 +105,12 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -U setuptools setuptools_scm wheel twine
+          pip install -U setuptools setuptools_scm wheel twine build
       - name: Build and publish
         env:
           TWINE_USERNAME: __token__
           TWINE_PASSWORD: ${{ secrets.TWINE_API_KEY }}
         run: |
           git tag
-          python setup.py sdist bdist_wheel
+          python -m build .
           twine upload dist/*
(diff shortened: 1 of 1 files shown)


Cryaaa:
Hey @jo-mueller,
again thanks for taking care of this. Shall I merge this before or after the next release is published?

jo-mueller:
@Cryaaa yes, definitely. the next release should then go through automatically.

Cryaaa:
> @Cryaaa yes, definitely. the next release should then go through automatically.

I'm not sure what you mean now :D just to confirm: merge this now and make the new release tag afterwards?

jo-mueller:
@Cryaaa Yes, this should be merged before the next release. If everything is already set for the next release then the release can also go through where I am concerned.

Cryaaa:
> > make the new release tag afterwards?
> 
> Please do #300 #301 #302 #303 before releasing. Thanks

Yes I wasn’t planning on releasing before the testing and documentation!

zoccoler (approved):

Issue #307, opened 2024-03-14
Avoid using Labels.color


zoccoler:
With napari 0.4.19, this warnings pops up after manually selecting a cluster

`FutureWarning: Labels.color is deprecated since 0.4.19 and will be removed in 0.5.0, please set Labels.colormap directly with an instance of napari.utils.colormaps.DirectLabelColormap instead.!`

It may be solved by replacing at least [this line](https://github.com/BiAPoL/napari-clusters-plotter/blob/5e937dc8c3347b6557a88d0fd730038182ec628b/napari_clusters_plotter/_plotter.py#L819), [this line](https://github.com/BiAPoL/napari-clusters-plotter/blob/5e937dc8c3347b6557a88d0fd730038182ec628b/napari_clusters_plotter/_plotter.py#L859) and [this line](https://github.com/BiAPoL/napari-clusters-plotter/blob/5e937dc8c3347b6557a88d0fd730038182ec628b/napari_clusters_plotter/_plotter.py#L877) with something like:

`layer_in_viewer.colormap= self.visualized_layer.colormap`

and

`"colormap": DirectLabelColormap(color_dict=cmap_dict),`

This needs to be tested.

gboscagli:
Hi all,
I'm using the Napari-SpatialData plugin (napari v.0.4.19.post1) obtaining the same Warning message. My data points in Napari are not colored, they remain uniformely grey. Is there a way to set the colormap on-the-fly or some other adjustments I can do? From #308 I saw that this remains an issue in Napari >=0.5.0.

zoccoler:
> Hi all, I know I'm probably a bit off-topic. I'm encountering the same Warning message using another Napari plugin (Napari-SpatialData). I have napari v.0.4.19.post1. My data points in Napari are not colored, they remain uniformely grey. To your knowledge, is there a way to set the colormap on-the-fly or some other adjustments I can do? From #308 I saw that this remains an issue in Napari >=0.5.0.

Hi @gboscagli !
This issue is because the Labels layer in napari>=0.5.0 does not have a `.color` attribute anymore. The napari developers left a deprecation warning so that plugin developers could update their plugins in case they were using that attribute.
In your case, it would be best to create the issue directly in the Napari-SpatialData plugin, so that developers there can work on the issue on their code.

gboscagli:
[...]

PR #308, opened 2024-03-26
Fix color future warning

zoccoler:
This is a small PR that addresses #307 and fixes a couple minor bugs


zoccoler:
If anyone can please give this an OK, I would merge it afterwards, no high-impact changes here

Cryaaa:
so in general everything LGTM but this only works for environments with napari >= 0.4.19 so there is a backward compatibility breaking change I guess we should mention?

zoccoler:
True, we would then have to update dependencies as well. I think we are moving in that direction for the next release, right?

Cryaaa:
That is true but maybe that would create issues with the devbio-napari system... @haesleinhuepf: do you know if relying on a specific napari version would cause issues for the devbio napari install?

zoccoler:
Let's wait for @haesleinhuepf feedback on this, it can be halted since it is not a major fix.

I would like to eventually have it implemented (maybe it can wait a couple months or so), because it does impact a little my usage of the current plotter by the [napari-flim-phasor-plotter plugin](https://github.com/zoccoler/napari-flim-phasor-plotter)

zoccoler:
I will break this PR into 2. 
- here, only changes needed to fix the `FutureWarning` will be kept
- the other minor changes were moved to #313 

This PR should remain open until we are sure we do not break `devbio-napari` installation.

ziw-liu:
As of napari 0.5.0 this is now an error:

```
TypeError: _add_layer_from_data received an unexpected keyword argument ('color') for layer type labels
```

It would be great if this fix gets into a release soon.

zoccoler:
OK now this is a bug with napari>=0.5.0.

The workaround until this is fixed is to use a lower napari version...

jni:
> this only works for environments with napari >= 0.4.19 so there is a backward compatibility breaking change

You can check the napari installed version and pass one key or the other based on that, which is what we did with napari-ome-zarr, here:

https://github.com/ome/napari-ome-zarr/pull/112/files#diff-3ac23664593f6255d26806df61d2364f3d3f0b66c53a4fad561a70877093239aR134-R139

Let us know if you have any questions about the new API!

zoccoler:
> You can check the napari installed version and pass one key or the other based on that, which is what we did with napari-ome-zarr, here:
> 
> https://github.com/ome/napari-ome-zarr/pull/112/files#diff-3ac23664593f6255d26806df61d2364f3d3f0b66c53a4fad561a70877093239aR134-R139


Oh nice @jni ! I will give this a try! Thanks!

zoccoler:
With the current state of this PR, there is no need to check for the napari version, it should work whether the `Labels` layer has the `.color` attribute or not because the colors dictionary is always converted with `DirectLabelColormap` (only available with napari>=0.4.18 TMK) into a colormap and assigned to the `.colormap` attribute.

I tried to install this version on top of the latest `devbio-napari` version (`0.10.1`) but I could not make it work when installing with `pip` becasue it brings npe>=0.7.0, 

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
devbio-napari 0.10.1 requires npe2<0.7.0, but you have npe2 0.7.7 which is incompatible.
```

This seems to be a requirement of napari>=0.5.1. For example, if I try to downgrade npe2 to 0.6.2, I get:

`- package napari-base-0.5.1-pyh9208f05_0 requires npe2 >=0.7.6`

So, napari-clusters-plotter **may** still work if installed with mamba/conda along with devbio-napari, but we can only test that after generating a new recipe in conda-forge (TMK).

I suggest we merge changes here regardless of the problems above because otherwise this is a bug of this plugin when napari>=0.5.0.
If this new version impacts devbio-napari installation, we should create an issue there and work on a solution from the outside, but not here.

@jo-mueller, if you agree, could you merge this?

jo-mueller:
Agreed @zoccoler, thanks for the fix!

jni:
> the colors dictionary is always converted with DirectLabelColormap (only available with napari>=0.4.18 TMK) into a colormap

It became available in 0.4.19. We should probably start adding `.. versionadded::` directives to docstrings when we add new things... 😅🙏

zoccoler:
> > the colors dictionary is always converted with DirectLabelColormap (only available with napari>=0.4.18 TMK) into a colormap
> 
> It became available in 0.4.19. We should probably start adding `.. versionadded::` directives to docstrings when we add new things... 😅🙏

You are right, my mistake! I had looked into napari API for other versions to check that. `DirectLabelColormap` only "shows up" in 0.4.19 (https://napari.org/0.4.19/api/index.html), thanks for the correction!

We have 0.4.19 as a requirement here (https://github.com/BiAPoL/napari-clusters-plotter/blob/44a658789fc119e339e320e4cab10238d67781eb/setup.cfg#L52C5-L52C19) so it should all be fine 😌

PR #311, opened 2024-03-28
Update doc layer types

zoccoler:
Here is my attempt to update the documentation and solve #301 .

2 particular points I would like some feedback:
- I am not sure how to suggest to derive the tracking features, currently I just load them and show how to manually classify them in a gif
- I did not find a way via plugins to load "points features" to the `Points` layer, that is why I created https://github.com/haesleinhuepf/napari-process-points-and-surfaces/issues/89, currently I am explaining how to add them via code


Cryaaa:
@zoccoler,
First of all thank you so much for taking care of this I love the examples you’ve added, very cool! As for the tracking data maybe @stefanhahmann can add a small explanation as to how the features were extracted? I think it was done with mastodon and I believe most tracking datasets would be generated by Fiji plugins but I’m also not very familiar with python/napari alternatives

zoccoler:
Hi Stefan @stefanhahmann , 
what do you think about the generation of the tracking data? Should I keep the description as it is?

zoccoler:
Thanks @stefanhahmann ! I added a link to the readme now with a brief explanation.

Cryaaa:
Awesome, then I don't see any issues with merging this!

PR #315, opened 2024-04-23
Add more tests for layers in 3 d/4 d

### LINKED ISSUE #302: Test current version before releasing

zoccoler:
This adds extra tests and should fix #302 

@jo-mueller , could you take a look?

### DIFF
--- a/napari_clusters_plotter/_tests/test_plotter.py
+++ b/napari_clusters_plotter/_tests/test_plotter.py
@@ -36,7 +36,7 @@ def get_labels_array():
     return label
 
 
-def test_plotting(make_napari_viewer):
+def test_plotter_on_labels2d(make_napari_viewer):
     viewer = make_napari_viewer()
     widget_list = ncp.napari_experimental_provide_dock_widget()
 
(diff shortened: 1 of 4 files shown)


napari_clusters_plotter/_tests/test_plotter.py
```diff
@@ -79,6 +79,278 @@ def test_plotting(make_napari_viewer):
...
+    z_slices = 3
+    n_timepoints = 2
+    n_labels = get_labels_array().max()
+    data_3d_timelapse = np.array(
+        [
+            np.roll(np.stack([get_labels_array() for _ in range(z_slices)]), t, axis=1)
+            for t in range(n_timepoints)
+        ]
+    )
+    # Create some random features
```
jo-mueller, line 202:
```suggestion
    assert len(data_3d_timelapse) == 4
    # Create some random features
```

napari_clusters_plotter/_tests/test_plotter.py
```diff
@@ -79,6 +79,278 @@ def test_plotting(make_napari_viewer):
...
+        data_2d_timelapse,
+        properties={
+            "label": label_column,
+            "feature1": feature1,
+            "feature2": feature2,
+            "annotations": annotations,
+            "frame": frame,
+        },
+    )
+
```
jo-mueller, line 163:
```suggestion
    assert len(data_3d_timelapse) == 4
```

jo-mueller (approved):
Just some minor suggestions to make sure that we are always handling 4D data

napari_clusters_plotter/_tests/test_plotter.py
zoccoler: I like that!

napari_clusters_plotter/_tests/test_plotter.py
zoccoler: Here it should be `data_2d_timelapse`. I will fix that.

jo-mueller (approved):

PR #325, opened 2024-07-01
Fix setting bin number manually

### LINKED ISSUE #320: Manual Histogram Bin Selection not Working
If we try to manually set the number of bins, the histogram does not get updated accordingly.

zoccoler:
This should fix #320


zoccoler:
I am merging this because this solves a bug in the current version.

PR #336, opened 2024-08-13
Bump version and automatic version determination

tsmejkal:
In the future you only have to update the version number in the init file instead of also setting it in the setup.cfg file.


jo-mueller:
Looks good to me!

jo-mueller (approved):

PR #345, opened 2024-09-17
unpin numpy version

Cryaaa:
Changed the dependencies to not rely on pinned numpy versions which are no longer supported by libraries like zarr. Thanks to @ziw-liu for bringing it up. Once merged this will close issue #215. I know this will be fixed anyway in [v 0.9.0](https://github.com/BiAPoL/napari-clusters-plotter/tree/v0.9.0) but it might be good to fix it in this release in case the new version has some backwards compatibility breaking changes.


ziw-liu:
Thanks for the fix! It would be great if this can get merged and release soon.

zoccoler:
Hi guys!

Just two questions to this. Could this enable numpy > 2.0 to be installed along with the plugin? If so, have you tested the plugin with it?

I am asking because numpy >2.0 brings many changes and some things may not work anymore, and the changes above wouldn't restrain it unless some other dependency does.

ziw-liu:
fwiw the CI tests for https://github.com/BiAPoL/napari-clusters-plotter/pull/345/commits/43f5f8297f41927eb5993b95092aaf9f90bb1583 was done with numpy>2:

```
numpy-2.0.2                |   py39h9cb892a_0         7.4 MB  conda-forge
```

zoccoler (approved):

PR #346, opened 2024-10-02
Add attribute for manual label opacity

thorstenwagner:
Hi,

when using the manual clustering method, the opacity of the labels is hard coded to 0.2. This is a problem for my usecase so I created an attribute.

Best,
Thorsten


thorstenwagner:
idk why CI fails, but I don't think my PR is the reason

zoccoler:
Hi @thorstenwagner ,

Sorry for the delay. This is a good idea! I will try one thing to check if it fixes the CI tests before I merge.

zoccoler:
Alright, this seems to work now, I am merging this PR. Thanks for the suggestion and patience @thorstenwagner !

Issue #383, opened 2025-04-16
Plugin not compatible with histogram2d data


zoccoler:
Many methods won't work in `0.9.0` if the active_artist is set to HISTOGRAM2D, mainly because they only exist for the SCATTER artist.

We should check in `biaplotter` that both classes have common methods so that transitioning form one artist to another runs smoothly.

zoccoler:
I think it would be best to match methods as much as possible to make the code smoother.

I started [this](https://github.com/BiAPoL/napari-clusters-plotter/pull/384) and [this](https://github.com/BiAPoL/biaplotter/pull/46) as attempts to solve this. Both are drafts because they are not fully working yet.

The histogram is important since some collaborators use it, so we should have it in `0.9.0` IMO

zoccoler:
That's great! Thanks!

I think the specific histogram features are not that many. Of course, we don't need all of them, I would consider connecting at least the histogram colormap, the log scale and the number of bins, which are available in the current version. The overlay colormap for non-categorical features could remain fixed for the time being (maybe `jet` or `viridis`, something other than and different from `magma`).

jo-mueller:
Hi @zoccoler , you mind if I close this issue?

PR #379, opened 2025-04-14
Escape empty selection

jo-mueller:
Relevant for #378 

Hi @Cryaaa , this solves your problem. In essence, the logic on the napari side is safe, the problem was here: The case of an empty selection (after a previous non-empty selection) is not escaped anywhere in the widgets, so that's a pretty serious flaw you found :)

I added specific `_clean_up` methods that handle the teardown of any remaining references to previous layers as well as escape statements that check whether there is data in the viewer before trying to do anything. I wasn't able to completely clean up the plotter itself (i.e., the plotted data), but that's something to fix over at the biaplotter (see https://github.com/BiAPoL/biaplotter/issues/45).

### Copilot description

This pull request introduces several changes to the `napari_clusters_plotter` package, focusing on enhancing the handling of empty layer selections, adding new functionalities, and improving test coverage. The most significant changes include the implementation of `_clean_up` methods, adjustments to handle empty input data, and the addition of new tests.

Enhancements to handling empty layer selections:

* [`src/napari_clusters_plotter/_algorithm_widget.py`](diffhunk://#diff-a47db6b3042af68f2254073d1a78dfd5ee32b1913eb50534391995aedadf2430R54-R60): Added `_clean_up` method to handle cases with no selected layers and updated `_on_update_layer_selection` and `_wait_for_finish` methods to incorporate these changes. [[1]](diffhunk://#diff-a47db6b3042af68f2254073d1a78dfd5ee32b1913eb50534391995aedadf2430R54-R60) [[2]](diffhunk://#diff-a47db6b3042af68f2254073d1a78dfd5ee32b1913eb50534391995aedadf2430R148-R153) [[3]](diffhunk://#diff-a47db6b3042af68f2254073d1a78dfd5ee32b1913eb50534391995aedadf2430R181) [[4]](diffhunk://#diff-a47db6b3042af68f2254073d1a78dfd5ee32b1913eb50534391995aedadf2430R207-R216)
[...]

PR #384, opened 2025-04-16
Clear plot and make it work with histogram

zoccoler:
Hi @jo-mueller ,

This PR should provide a way to clear the plot after [this PR](https://github.com/BiAPoL/biaplotter/pull/46) gets merged in biaplotter.

However, since some methods that both the scatter and the histogram artists would use are not the same, there is work in progress here to also make them compatible, potentially solving #383.

### DIFF
--- a/src/napari_clusters_plotter/_new_plotter_widget.py
+++ b/src/napari_clusters_plotter/_new_plotter_widget.py
@@ -6,7 +6,10 @@ import numpy as np
 import pandas as pd
 from biaplotter.plotter import ArtistType, CanvasWidget
 from matplotlib.pyplot import cm as plt_colormaps
-from nap_plot_tools.cmap import cat10_mod_cmap, cat10_mod_cmap_first_transparent
+from nap_plot_tools.cmap import (
+    cat10_mod_cmap,
+    cat10_mod_cmap_first_transparent,
+)
 from napari.utils.colormaps import ALL_COLORMAPS
 from qtpy import uic
 from qtpy.QtCore import Qt, Signal
@@ -186,11 +189,17 @@ class PlotterWidget(BaseWidget):
 
         # check hue axis for categorical data
         if self.hue_axis in self.categorical_columns:
-            if active_artist == self.plotting_widget.artists[ArtistType.SCATTER]:
+            if (
+                active_artist
+                == self.plotting_widget.artists[ArtistType.SCATTER]
+            ):
                 self.plotting_widget.active_artist.overlay_colormap = (
                     cat10_mod_cmap
                 )
-            elif active_artist == self.plotting_widget.artists[ArtistType.HISTOGRAM2D]:
+            elif (
+                active_artist
+                == self.plotting_widget.artists[ArtistType.HISTOGRAM2D]
+            ):
                 self.plotting_widget.active_artist.overlay_colormap = (
                     cat10_mod_cmap_first_transparent
                 )
@@ -372,7 +381,15 @@ class PlotterWidget(BaseWidget):
             if layer.name not in self._layer_colormap_cache:
                 if isinstance(layer, napari.layers.Labels):
                     self._layer_colormap_cache[layer.name] = layer.colormap
-                elif isinstance(layer, (napari.layers.Points, napari.layers.Shapes, napari.layers.Vectors, napari.layers.Surface)):
+                elif isinstance(
+                    layer,
+                    (
+                        napari.layers.Points,
+                        napari.layers.Shapes,
+                        napari.layers.Vectors,
+                        napari.layers.Surface,
+                    ),
+                ):
                     self._layer_colormap_cache[layer.name] = layer.face_color
 
         # Insert 'MANUAL_CLUSTER_ID' column if it doesn't exist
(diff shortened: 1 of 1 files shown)


src/napari_clusters_plotter/_new_plotter_widget.py
```diff
@@ -219,6 +238,13 @@ def _on_frame_changed(self, event: napari.utils.events.Event):
             self.plotting_widget.active_artist.alpha = alpha
             self.plotting_widget.active_artist.size = size
 
+    def _clear_plot_data(self):
+        """
+        Clear the plot data.
+        """
+        self.plotting_widget.active_artist.data = None
+        self.plotting_widget.active_artist.color_indices = None
+
```
jo-mueller, line 247: Can you make it compatible with the implementation in [this branch](https://github.com/BiAPoL/napari-clusters-plotter/tree/escape-empty-selection)? I also implement a similar logic there because a few more things need to be considered in case of an empty selection.

```Python
    def _clean_up(self):
        """In case of empty layer selection"""

        # disconnect the events from the layers
        for layer in self.viewer.layers.selection:
            layer.events.features.disconnect(self._update_feature_selection)

        # reset the selected layers
        self.layers = []

        # reset the selectors
        for dim in ["x", "y", "hue"]:
            selector = self._selectors[dim]
            selector.blockSignals(True)
            selector.clear()
            selector.blockSignals(False)
```

src/napari_clusters_plotter/_new_plotter_widget.py
```diff
@@ -170,25 +177,37 @@ def _replot(self):
...
+                self.plotting_widget.active_artist.overlay_colormap = (
+                    cat10_mod_cmap
+                )
+            elif (
+                active_artist
+                == self.plotting_widget.artists[ArtistType.HISTOGRAM2D]
+            ):
+                self.plotting_widget.active_artist.overlay_colormap = (
+                    cat10_mod_cmap_first_transparent
+                )
```
jo-mueller, line 205: Instead of the long `if/else` statement, maye it would be easier to create a dictionary like:

```python
colormap_dict = {
  ArtistType.SCATTER: cat10_mod_cmap,
  ArtistType.HISTOGRAM2D: cat10_mod_cmap_first_transparent
}
```
Here, the colormap could be set simply by

```
self.plotting_widget.active_artist.overlay_colormap = colormap_dict[type(active_artist])
```

and keep the code a bit more concise?

src/napari_clusters_plotter/_new_plotter_widget.py
```diff
@@ -331,32 +357,51 @@ def _on_update_layer_selection(
...
+                    (
+                        napari.layers.Points,
+                        napari.layers.Shapes,
+                        napari.layers.Vectors,
+                        napari.layers.Surface,
+                    ),
+                ):
+                    self._layer_colormap_cache[layer.name] = layer.face_color
+
+        # Insert 'MANUAL_CLUSTER_ID' column if it doesn't exist
```
jo-mueller, line 395: Is the idea here to cache the color values for a particular layer in case a large number of items needs to be colored? I think it's a great feature to have but I'm not sure I'd implement it like that. Maybe we bring this feature to a different PR?

src/napari_clusters_plotter/_new_plotter_widget.py
```diff
@@ -456,20 +513,40 @@ def _reset(self):
...
+            (
+                napari.layers.Points,
+                napari.layers.Shapes,
+                napari.layers.Vectors,
+                napari.layers.Surface,
+            ),
+        ):
+            layer.face_color = cached_colormap
+        layer.refresh()
+        return
```
jo-mueller, line 548: This will work only for `Points` data. For Tracks/Surfaces/Vectors/Shapes the properties have different names (`track_colors`, `vertex_colors`, `edge_colors`). That's why I added the `if/elif`s a bit further downn to overwrite the correct property for each kind of layer.

In this case, I think it would be easier to call the function with the cached colormap (is `cached_colormap` an `Nx4` rgba array?) in the first place rather than passing it as an extra argument. My idea behind having this function was to separate the logic that applies the colors (`_apply_layer_color`) from the functionality that parses the input on the plotter side (`_color_layer_by_value`).

That way, the `_apply_layer_color` function could be used elsewhere for a bit of a unified access to layer coloring since it's so different for all the different layer types :/

jo-mueller (changes requested):
Hi @zoccoler , thanks for the PR! I think my major comment would be that this PR brings two features - cleaning up the plot correctly, applying artist-type specific colormaps (feature no 1) as well as caching colormaps for re-use (feature no.2)?

I think I agree with the implementation of feature no. 2 but I'm not so sure about the second one (see comments). 

Parts of this I also implemented [here](https://github.com/BiAPoL/napari-clusters-plotter/pull/379) (escaping no layer selected). I think we came up with the same idea (writing a clean-up function) that executes in case no layer is selected. I also added this function for the base widget of the plotter as well as the `DimensionalitReductionWidget` and the `ClusteringWidget`, as they'll suffer from the same issue (weird stuff happening with no layer selected).

On a scale from 1-10 (very hard) - how hard would it be to take this PR apart into separate PRs for the separate features?

src/napari_clusters_plotter/_new_plotter_widget.py
zoccoler: Not sure what you mean, sorry 😅

src/napari_clusters_plotter/_new_plotter_widget.py
zoccoler: Sure!

src/napari_clusters_plotter/_new_plotter_widget.py
zoccoler: Unfortunately yes. I don't like storing a whole colormap for every layer either, but the way the colors are generated for the Labels layer in napari may vary (they say it is random in the documentation). It is sort of consistent for the first 'x' labels, but it may change depending on how many labels the layer has. We could bring it to a different PR, sure. The problem arouse because:
- I selected some points in the plot, got the layer colored
- I cleared that selection (chose color `0` in the plot), the labels in the layer got all light gray (for scatter) or the layer got fully transparent (for histogram)

This is obviously because the color `0` from the corresponding default colormap is being used to color all labels. I don't like that behaviour, one could never go back to the original colors if we do not provide a way to do so.

src/napari_clusters_plotter/_new_plotter_widget.py
zoccoler: Indeed I did not test for layers other than `Labels` 😅 
I need to test more to answer this.

jo-mueller:
Hi @zoccoler do you mind if I close this? I think this functionality has been brought to the 0.9.0 branch in previous PRs of you and me :)

zoccoler:
Sure, I am closing it.

PR #390, opened 2025-04-25
Connect histogram optional features

zoccoler:
Hi,

I started connecting events for the Advanced tab. For that, I updated the .ui file a little bit.

This is still a draft PR (WIP), but I wanted to expose it more explicitly to share this starting point.

At this point, I connected the log checkbox and the overlay_colormap Combobox (the other widgets start hidden from the interface for now).

To Do:
- [x] connect histogram colormap ComboBox
- [x] connect manual bins SpinBox
- [x] connect Auto PushButton
- [ ] connect Hide non-selected clusters CheckBox
- [x] perhaps connect napari `layer.events.visible` to [`hide_color_overlay` method](https://github.com/BiAPoL/biaplotter/blob/233336f1cf184da31358d9eac5f241ec15dee929/src/biaplotter/plotter.py#L394C9-L394C27) and vice-et-versa


zoccoler:
Hi @jo-mueller ,

This is ready for review. It needs the main development branch of the biaplotter (v0.3.0, not yet release but ready).

 I connected all the options in the advanced tab, except the hide non-selected clusters checkbox.

The functionality of painting the selected layer(s) with the colors drawn in the plotter and resetting the colors upon unselecting is working, at least for the Labels layers. Please give it a try.

jo-mueller:
Hi @zoccoler I only could have a quick look at it on my smartphone so my insight is limited. One idea that came to me though is the following:

From what I understand there are a bunch of widgets that turn visible/invisible depending on whether the HISTOGRAM2D is enabled. You could put all the widgets for the histogram in a separate container (I think there's something like a QFrame or QWidgetGroup) so you'd only have to turn one object visible and save yourself a bit of code.

zoccoler:
Hi @jo-mueller ,

Thanks for checking this up so quickly. That would save just a few lines of code because just 2 containers are histogram specific: bins and histogram overlay colormap.

There are other conditions that do not related to hide/show, for example:

- Bins: setting the auto bins should, in my opinion, set the value in the manual bin spinbox and disable it. By the way, I took the "Set" button out of the interface.

- Log checkbox: it must always be available. It can be used by the scatter plot to display a non-categorical hue colors in log, and in all cases for the histogram.

I would like to further simplify the logic, but at the moment, I focused on getting it working. One thing that I find confusing is that some widgets callbacks arw connected to the "plotter_needs_update" signal, which calls _replot, others are directly linked to specific callback functions that eventually call _replot, and there are some methods unused that we could remove

zoccoler:
The most intricated logic to me at the moment is the one that colors the layer depending on different configurations.

zoccoler:
I noticed that changing some settings, especially those related to the color_overlay, would bring the overlay back on top (sometimes permanently) regardless of the `show_color_overlay` property (reflecting the ToggleButton state). This issue was coming from the biaplotter.

[This PR in the biaplotter repo](https://github.com/BiAPoL/biaplotter/pull/51) should fix it, making the color overlay visibility fully determined by the `show_color_overlay` property. So after biaplotter `0.3.1` release, the recent changes here should also work (minimal biaplotter version already pinned in this PR now).

Below is a small demonstration of the following behaviors:
- If layer(s) is/are selected, then overlay colors apply (given show_overlay button is checked - eye icon is visible)
- If layer(s) become unselected, default colors of the unselected layers are restored
- Toggling the `show_color_overlay` property (eye button) shows/hides the color overlay over the plot (histogram/scatter) and also applies/restores colors to selected layer(s) 

The code and gif below were made on napari `0.5.6`.

ncp_ff

zoccoler:
I just solved minor conflicts to get tests running, sorry...

`biaplotter` `0.3.1` is out 🎉

zoccoler:
Oh, I will take a look into #395

jo-mueller:
@zoccoler The new bug you found in #395 I could also reproduce here. I think I found the issue, will fix it directly here on this branch.

jo-mueller (approved):

PR #422, opened 2025-06-04
allow mixed selections

### LINKED ISSUE #331: Meta layer navigation widget?
When the clusters plotter supports cross-layer clustering functionalities, we should provide users with a small widget that helps to navigate layers, especially if there are many of them. Such a widget could offer the following functioanlity:

- Select all layers of a particular kind (all `Labels`/`Surface`/etc layers, for instance)
- Select all layers as defined by a searchstring for their layer name
- Functionality for putting multiple layers in a grid layout (similar to what's already in napari), but put linked layers in the same place - how to interact with linked layers?

jo-mueller:
Fixes #331 

This PR makes some changes to how unsupported layers are treated upon selection. In essence, this PR changes this so that unsupported layers (e.g., `Image`) are simply ignored. Previously, if an unsupported layer type is encountered in the selectionn/drawing process, the functions return and aborts so that nothing can be drawn. With this PR, a the clusters-plotter functionality is simply executed on the layers that are supported while the others are ignored.

This can be especially helpful if layers arent't organized as stringently as in the `bbbc1` example dataset, where all the labels are on top and all the images are on the bottom. I just learned, that the napari grid view provides the `stride` argument, which groups `n` layers together in a single grid tile, depending on the setting of the `stride` parameter. This saves ourselves the trouoble of setting `translate` values for the individual images in the `bbbc1` dataset.


zoccoler:
This works for the plotter, but not when opening the other widgets, only after re-selecting layers:

06 06 2025_17 20 49_REC

jo-mueller:
Fixed in 4b293d92f5efae3b4c33eb466b2cfe092dd61730

PR #430, opened 2025-06-06
replace calls to `_replot()` by signal emission and fix plot clearing

jo-mueller:
This PR makes sure that the plot is properly erased after clicking the clear-button. It also replaces the calls to the `_replot` function by a signal emission, which *then* triggers the replot. This has the advantage of decoupling the plotting from the function that calls it, which should make the experience a bit more smooth.


zoccoler:
This did not work for me yet:

06 06 2025_17 15 31_REC

PR #366, opened 2025-02-27
add LAYER_SELECTED_DATA_CLUSTER_ID to cluster from layer space

ClementCaporal:
Hello,

Attempt to close #363 

In this PR I propose to add to `v.0.9.0` the point layer features column `LAYER_SELECTED_DATA_CLUSTER_ID`.

The event connection is added during `_on_update_layer_selection` (and never removed?). This allow the user to select points in the layer to appear on the plotter canvas through features.

---

`v0.9.0` seems amazing! For example, the ability to select several layers and to plot them on the same plotter canvas is something I didn't know I needed.

I feel I didn't understand how to change the `color_indices` by using another `CLUSTER_ID` than `MANUAL_CLUSTER_ID` but maybe it is not implemented yet?

I tried to keep the PR minimal because I am unsure about how v0.9.0 will be developed, but I can of course change the implementation / names.

Thank you for your time

PR #441, opened 2025-06-19
connect napari-sided object selection in plotter widget

ClementCaporal:
Follow-up on https://github.com/BiAPoL/napari-clusters-plotter/pull/366

---

Apologies for the delay in getting back to you!

Since I couldn't copy-paste your comments directly, here's a summary of what I’ve addressed based on your feedback:

- I incorporated `your _is_selectable` and `_get_selected_objects` as private functions and applied them where needed.
-  I updated the test to check the color indices directly from the current plotter.
- Regarding the variable name `_VIEWPORT_CLUSTER_ID`: I kept the original name `SELECTED_DATA_LAYER_CLUSTER_ID` to stay consistent with the naming used in the napari layer selected_data. I don't have a strong opinion on this, this is just a proposition, I can change it.

---

This PR is not ready yet (and it doesn't pass all the tests) because : 
I’m still facing an issue related to selection on the labels layer that I think are out of scope of this PR.

To handle label selection, we need to listen to the `selected_label` event:
https://github.com/napari/napari/blob/86e6d27c547f3fad22ddfe8835c4869d3f0931ed/src/napari/layers/labels/labels.py#L382

From there, we update the layer features and emit `self.plot_needs_update`. This eventually leads to updating `layer.colormap`, which unfortunately triggers the `selected_label` event again:
https://github.com/napari/napari/blob/86e6d27c547f3fad22ddfe8835c4869d3f0931ed/src/napari/layers/labels/labels.py#L571

This creates an unintended recursive loop.

I _think_ this might be a bug on napari’s side (it's unclear to me why changing the colormap should emit a selected_label event). Unless I’ve missed something, I’m will open an issue on their repo based on your review.

Thanks again for your time and for the excellent plugin!


src/napari_clusters_plotter/_new_plotter_widget.py
```diff
@@ -745,3 +773,39 @@ def _reset(self):
...
+        Retrieve id of selected object on napari canvas"
+        """
+        if isinstance(layer, napari.layers.Points):
+            return list(layer.selected_data)
+        elif isinstance(layer, napari.layers.Labels):
+            return [layer.selected_label]
+        else:
+            raise TypeError(
+                f"Layer type {type(layer)} is not supported for selection."
+            )
```
jo-mueller, line 796: This doesn't necessarily have to be a member of the `PlotterWidget` - none of the functionality relies on `self`. Originally, some other functionality in here (e.g., `_set_layer_color`) weren't member functions but I think that was lost somewhere along the way. 

Still - moving it to `_utilities.py` could help to keep this widget a bit cleaner here :)

src/napari_clusters_plotter/_new_plotter_widget.py
```diff
@@ -522,6 +526,30 @@ def _on_update_layer_selection(
...
+        Update the layer selected_data to feature.
+        """
+        selected_data = self._get_selected_objects(layer)
+        cluster = np.zeros(len(layer.features))
+        cluster[list(selected_data)] = 1
+        # set categorical to be selectable in "Hue" dropdown
+        layer.features["SELECTED_DATA_LAYER_CLUSTER_ID"] = pd.Categorical(
+            cluster
+        )
+        self.plot_needs_update.emit()
```
jo-mueller, line 551: Not sure - does it make sense to iterate over all the layers and collect the selected objects inside this function? That way, you could save yourself the `lambda` callback connection above. Somethhing like this:

```python
for layer in self.layers:
  selected_data = self._get_selected_objects(layer)
  cluster = np.zeros(len(layer.features))
  cluster[list(selected_data)] = 1
  # set categorical to be selectable in "Hue" dropdown
  layer.features["SELECTED_DATA_LAYER_CLUSTER_ID"] = pd.Categorical(cluster)
```

jo-mueller (changes requested):

jo-mueller:
Hi @ClementCaporal ,

just gave it a try, it looks pretty much how I envisioned it to work for selectable objects :)

>This creates an unintended recursive loop.

I was wondering why it was so slow for labels 😅 

> From there, we update the layer features and emit self.plot_needs_update. This eventually leads to updating layer.colormap, which unfortunately triggers the selected_label event again:

Maybe we could just mute/block this event while updating the colormap? I just tried to add the following into the `_set_layer_color` function and it worked:

```
        elif isinstance(layer, napari.layers.Labels):
            from napari.utils import DirectLabelColormap

            from ._utilities import _get_unique_values

            # Ensure the first color is transparent for the background
            colors = np.insert(colors, 0, [0, 0, 0, 0], axis=0)
            color_dict = dict(zip(_get_unique_values(layer), colors))
            layer.events.selected_label.block()  # avoid triggering the infinite loop
            layer.colormap = DirectLabelColormap(color_dict=color_dict)
            layer.events.selected_label.unblock()  # restore functionality for whatever purpose it serves
```

